PR #5, opened 2017-06-11
Cancel playlist when exiting Gameplay

Nickito12:


martensm:
Is `ScreenGameplay`'s destructor the right place to do this?

Part of this is already done process is already done here:
https://github.com/Nickito12/etterna/blob/ffa358b1ade6142560ff7bd5db1ef3dfeda80f32/src/ScreenGameplay.cpp#L2262

Wouldn't adding
```cpp
SONGMAN->playlistcourse = "";
```
in the `if` block on line 2262 accomplish the same thing?

Nickito12:
Moving this to https://github.com/etternagame/etterna/pull/6

PR #6, opened 2017-06-13
Stop playlist if gameplay is left early

Nickito12:


martensm:
Could make a simple function to avoid duplicating this check but I don't think it's that important. Other than that, LGTM.

Feel free to merge @MinaciousGrace.

PR #16, opened 2017-07-17
Clipboard stuff

Nickito12:
Add getclipboard function to lua and make it so the online chatbox can paste it on ctrl+v


Themes/Til Death/BGAnimations/ScreenSelectMusic decorations/songsearch.lua
```diff
@@ -1,10 +1,11 @@
 local searchstring = ""
-local englishes = {"a", "b", "c", "d", "e","f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t", "u", "v", "w", "x", "y", "z",";"}
+local englishes = {"?","-",".",",","1","2","3","4","5","6","7","8","9","0","a", "b", "c", "d", "e","f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t", "u", "v", "w", "x", "y", "z",";"}
```
martensm: Please allow this to support all printable Unicode characters or at least more than this set of characters. Non-english titles can't be searched as a result of this.

PR #39, opened 2017-08-02
Make portable.ini unnecessary

despoa:
Functions that make folders created by the Etterna program such as Cache and Save be in local user directories have been removed as well.


martensm:
This limits flexibility in regards to where Etterna will save files. If this functionality is removed, a substitute is necessary.

PR #54, opened 2017-09-01
Removed the MinaCalc from the root and rebuilt the calc to work on x64

TheROPFather:
This seems to fix the Linux build completely - I'm not 100% certain how it works but it does, this allows users on 64bit hosts to fully compile the program with no flags required.


martensm:
Will merge for now since there is mutiple reports of this working but Travis failure should be investigated. There is possibly an issue with the `binutils` version -- can this be worked around?

PR #56, opened 2017-09-02
Image cache

Jousway:
change BannerCache into ImageCache and added Background and CDTitle support
Added Lua calls into Sprite so people can use LoadFromCached("ImgType", "ImgPath")


Jousway:
changing this into Load() so that it doesnt break backwards compatibility and caches on default

Jousway:
I'm to lazy to write a proper ver atm so I'll keep it in my own ver for now

PR #57, opened 2017-09-03
Adds initial vanilla Lua(5.1.5) support

caiohsr14:
In order for future LuaJIT testing.


martensm:
LGTM, tested it and it seems to work fine. @MinaciousGrace should probably review this as well.

Thanks for the PR!

PR #62, opened 2017-09-05
Changing BannerCache into ImageCache

Jousway:
Basicly removes the BannerCache and replaces it with a system that supports more images

The default supported images are
- Banners
- Backgrounds
- CDTItles
- Jackets
- CDImages
- Discs

---

Adds a new line to Themes called
`ImageCache=`

example usage:
`ImageCache="Banner,Background,CDTitle"`

---

Adds new Lua call to sprite called
`LoadFromCache();`

example usage:
```
Def.Sprite{
    OnCommand=function(self) self:LoadFromCache("sType", sPath);
}
```


Jousway:
I change RString to std::string inside ImageCache
Had to add some extra code like
```
Line 337: RString otImagePath = sImagePath; // Remove this when Global RString to std::string convert.
Line 374: RString otImagePath = sImagePath; RString otError = sError; // Remove this when Global RString to std::string convert.
```
Because they accept only RString, When everything is converted to std::string we can remove these.

src/Song.cpp
```diff
@@ -1420,10 +1383,25 @@ RString Song::GetCacheFile(RString sType)
...
+		return PreDefs[sType.c_str()];	
+	
+	// Get all image files and put them into a vector.
+	vector<RString> song_dir_listing;
+	FILEMAN->GetDirListing(m_sSongDir + "*", song_dir_listing, false, false);
+	vector<RString> image_list;
+	vector<RString> fill_exts = ActorUtil::GetTypeExtensionList(FT_Bitmap);
+	for( std::string Image : song_dir_listing )
+	{
+		std::string FileExt = GetExtension(Image).MakeLower();
```
martensm: `transform(begin(Image), end(Image), begin(Image), ::tolower);`

src/PrefsManager.h
```diff
@@ -57,26 +57,16 @@ enum ShowDancingCharacters
 	NUM_ShowDancingCharacters,
 	ShowDancingCharacters_Invalid
 };
-enum BannerCacheMode
+enum ImageCacheMode
```
martensm: Using an enum class makes this more type safe but it's not that important.

Jousway: most of the stuff in prefs use it so I'ma leave it in for now, if we are planning to remove it entirly lets remove it then 👍

Nickito12:
Is this 100% merge-ready?

PR #66, opened 2017-09-09
Add custom timing windows eval screen rescoring

caiohsr14:
These should be added to the Scripts/01 timing_windows_config.lua file in order to work(there are a few examples there too). The previous judge rescorer at eval screen has also been moved to lua(so it's not disabled anymore) and a menu option for using the custom timing windows has been added into the player options menu.
Also added a command to reset the current displayed score to the original one, currently bound to MenuUp.


caiohsr14:
@MinaciousGrace change lines 390 and 413 from scores.lua, `or 0.0` to `or -1.0`

PR #67, opened 2017-09-09
Net stuff, Scripts moved to fallback and delete empty scripts

Nickito12:
* Some net screen fixes(From updates to offline screens and lua 5.1 syntax)
* Delete empty lua files and make the installer delete til death
* Move some useful Scripts from Til Death to fallback


stepmania.nsi
```diff
@@ -410,6 +410,7 @@ Section "Main Section" SecMain
 
 	; remove and install themes
 	RMDir /r "$INSTDIR\Themes\_fallback"
+	RMDir /r "$INSTDIR\Themes\Til Death"
```
martensm: Can you make a separate commit for this? I don't want to do this here because it's something that needs to be discussed further.

Themes/Til Death/BGAnimations/ScreenEvaluation decorations/offsetplot.lua
```diff
@@ -1 +0,0 @@
---moved to ../
```
martensm: Shouldn't this be in the previous commit?

Themes/Til Death/BGAnimations/ScreenEvaluation decorations/wifeScold.lua
```diff
@@ -1 +0,0 @@
--- gone
```
martensm: Shouldn't this be in the previous commit?

martensm:
There is too much going on in this PR, it needs to be more focused. If you just PR the scripts being moved to fallback changes then i'll merge it. I'd suggest creating a separate PR for each bullet point so they can be tracked more easier.

Nickito12:
Ok, the net stuff wa re-PR'd in https://github.com/etternagame/etterna/pull/86
I'm gonna close this

PR #70, opened 2017-09-12
Added background quad to ScreenStatsOverlay for FPS/AV FPS and VPS

Jousway:
how is this guys, here is an example


Jousway:
`actually can you check if this works ok with different resolutions and aspect rations
i'm concerned about hardcoded 80`

it should work fine on any resolution here is a pic of it at 640x480 4:3

martensm:
Thanks for checking that. LGTM

PR #75, opened 2017-09-17
Added a global StringConvert function

Jousway:
Seeing that RString is a mess and we want future commits to use std::string, some older code depends on RString, we fix that by using this convert.

### DIFF
--- a/src/global.cpp
+++ b/src/global.cpp
@@ -55,6 +55,18 @@ void NORETURN sm_crash( const char *reason )
 #endif
 }
 
+std::string StringConvert(RString String)
+{
+	std::string rString = String;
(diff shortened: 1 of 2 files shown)


Nickito12:
Im not sure i see much point to this, iirc this kind of thing can be done manually and if a function requires an RString specifically it can just be given a string.c_str() if i recall correctly. If a function takes an RString& (Referecen) i dont think this function is useful (And the only way to go is either refactor said function or have an RString)

martensm:
I've decided against merging this. Mostly for the reasons @Nickito12 pointed out and because I don't think it's actually going to help much for migrating to `std::string`.

PR #79, opened 2017-09-23
Adding more Noteskin features.

Jousway:
Added "Overlay Receptor" which is above tapnotes.
Added "Underlay Receptor" which is under tapnotes.
Added this because it uses "Receptor" as fallback and "Receptor" shouldnt be overwritten.

Added "Held Explosion Dim" and "Held Explosion Bright".
Makes it easier for users to define which explosion they want after the hold ended.

At the moment it only supports Dance Single, Can add more support if requested.


Jousway:
do want to note this doesnt break compatilibity with old noteskins

Nickito12:
Is this 100% merge-ready?

PR #82, opened 2017-10-07
libCURL

Nickito12:
1. libCURL
1. Profile Uploads
1. Pack Downloads (MULTI curl interface, generic and gameplay specific max dl speeds in preferences, supports multiple downloads, packlist downloaded from a url in preferences(Only allows downloads from lua from this packlist), in JSON format as follows:
{{"name"="packname", "size"=0, "url"="https:etc"}, {"name"="packname2", "size"=10000, "url"="https:etc"}}
, downloads http/s urls passed as arguments to the program, deleted sm filetransfers)
1. Fix DifferentialReload for smzips(Flushes)
1. Refactor a bit ScreenTextEntry(Now works with nested screens of textentry since its not always using the globals)
Todo:
- [x] Make sure libcurl is properly found on mac/linux
- [ ] Make the code cleaner
- [ ] Use smart pointers
- [ ] Make sure the lua interface works (Highest suspect is Download lua table object)
- [ ] Add packs tab in til death to online screens


martensm:
Non-exhaustive list of things to fix:

* don't use raw pointers, use `std::unique_ptr` as private members of the classes
move `#if !defined(WITHOUT_NETWORKING)` to top of header file
* remove commented out code
* make `DownloadManager *DLMAN = NULL`; a `std::shared_ptr`. i know that this isn't how stepmania operates with other singletons.
* use `nullptr` instead of `NULL`
* use uniform initialization instead of most of the code that's current in the constructor for setting primitive values (see `RageSoundReader_MP3.h` for an example)
* formatting is all over the place, lots of spurious newlines
* `UpdateAndIsFinished` is massive, can you refactor this?
* don't return a pointer to a vector from `GetPackList`, i don't see the point of doing this instead of just returning it by value
* don't use `RString`

PR #84, opened 2017-10-09
sqlite (WIP)

Nickito12:
1. Link sqlite
1. Move xml profile logic to xmlprofile
1. DBProfile
1. Only write PBs to etterna.xml(Its write only now, we use the DB to store profiles for ourselves)
1. PB-only db option


Nickito12:
Moved to https://github.com/etternagame/etterna/pulls

PR #89, opened 2017-10-19
ignore vs2015+ cache/settings dir

lauriys:
there is no way this can screw up again


lauriys:
please forget i exist

PR #96, opened 2017-10-27
Init MESSAGEMAN earlier

hexpointer:
It is undefined behaviour to call MESSAGEMAN->Broadcast before initialization (compilers are free to compile away the null check on MessageMan.cpp:209, and in fact g++ does, leading to a SIGSEGV on startup).


MinaciousGrace:
vanilla has had this merged for a couple years and i didn't have any issue when testing it so, ok

PR #99, opened 2017-11-14
Downloads stuff

Nickito12:
Fix profile upload crash for empty profiles, gameplay dl speed throttling during gameplay, use a map for dlman's downloads (urls as key), add http::// if url doesnt have it already and make dlman properly reload songs when downloads are finished during gameplay


Nickito12:
Cherry picked into https://github.com/etternagame/etterna/pull/82

PR #101, opened 2017-11-15
Numberinos

Nickito12:
Make songsearch only input numbers if ctrl is pressed, and dont change tab if it is pressed


Nickito12:
Closing this since https://github.com/etternagame/etterna/pull/102 makes this useless

PR #121, opened 2017-12-04
luaJIT

Nickito12:
Only supports windows (Requires adding mac and linux .a libraries to extern/LuaJIT, and possibly renaming the file names in StepmaniaCore.txt for them)


Nickito12:
Moved to branch expEverything

PR #134, opened 2018-01-05
sqlite song cache

Nickito12:
- Link Sqlite and sqlitecpp
 - Move xml profile logic to xmlprofile.cpp (And add an unused dbprofile prototype)
 - Replace song cache with 1 sqlite db file
 - Change song load logic : Load cache entries -> Check each cache entry's validity and finish loading (Or not) -> Differential reload songs and additionalSongs
 - Turn ImageCache default value to Off
 - New preference to not use directory hashes in cache load validity check
 - Turn a lot of SM/SSC loader functions into static ones(With member wrappers) so they can be used without a loader object when saving cache entries

Things that should probably be done before merging:

 - Replace vectors and MSD values being stored as strings/text in the db with the actual values and a relational model (Should be both faster and make things more clear/easier to read)
 - Find out why theres a diff in ScoresAtRate::CreateNode and probably remove it from the PR
 - Look into improving loading window messages to provide more info


Nickito12:
Moved to branch expEverything

PR #135, opened 2018-01-05
Fix multiple profiles

Nickito12:
Store profile directory in profile as m_sProfileID and use that as key for maps in scoreman


Nickito12:
Moved to branch expEverything

PR #142, opened 2018-02-08
Added 50% and 45% Distant Modifier

TheKamilia:
Add this to fallback too


Nickito12:
I'm not sure partial perspectives are a good idea in default options. Imo this should be something players can do themselves (And, ideally, a theme can have functionality to choose any % for a perspective, but having 100XperspectiveCount items in the options doesnt strike me as a good idea)

TheKamilia:
I use a laptop to play and adding % on persp are difficult and incovenient to the average player, this is a good way to use it

PR #143, opened 2018-02-08
Added a Update Option

TheKamilia:


Nickito12:
This isn't very clear. Is it to add an "Update" item in the title screen that opens the browser in EO's download page?
This *might* be a good idea. I'm not fully convinced. There should also be discussion as to weather this should link to EO or to the GH's releases page

TheKamilia:
Idea taken from SM 1.64, requested by someone in my stream

Nickito12:
I feel like if we wanted to do this we should make it check if it actually needs to update (We could use http requests to somewhere that would hold the last version, and compare), and then either offer a download link OR the ability to update ingame (Probably the first one considering the work it'd take)

Nickito12:
If youre still interested, in the branch "expRelease" there's this (https://github.com/etternagame/etterna/blob/expRelease/src/DownloadManager.cpp#L1185). Basically DLMAN requests the last version from the server in preferences (Defaults to EO) as a string on startup. I think you could compare that to the current version (You can see an example of how to get it in til death/bganimations/_PlayerInfo.lua iirc or something like that) to see if displaying something like "Old version, needs updating" is needed. You can contact me in discord if you want help, you can find me in the EO or EDG discord servers.

nico-abram:
I'm closing this for now since it's been over two months with no response

PR #148, opened 2018-02-22
Recompile discord RPC and LuaJIT with -fPIC for static lib.

TheROPFather:
Recompiled discord lib with clang and enabled -fPIC. Recompiled luaJIT with clang and also enabled -fPIC.


Nickito12:
:+1:

PR #155, opened 2018-03-31
 Add a player tag system to Til Death

caiohsr14:
Allows the player to add tags to stepcharts (not songs), filter by them and removing them as well.

Usage example: https://puu.sh/zT2wH/c7f66103c1.webm

The tags are stored in such a way that allows easy sharing from player to player, since what it generates is a config file under `Save/Til Death_settings` called `tags.lua`, which contains a readable table with tags and their respective chart keys:

image


Nickito12:
:+1:

PR #156, opened 2018-04-07
deleted character folder

TheKamilia:
seriously has anyone used this in the past 10 years?


nico-abram:
I think some people use this

PR #163, opened 2018-05-16
proposed sample accurate metronome based sync track

tylerstraub:
Based on a conversation in the EtternaDev Discord, we discovered that the existing file (`_sync music.ogg`) was inaccurate. The transients in the drum pattern were roughly 44ms behind. This music was created using ProTools and based around a sample accurate metronome. Also contains an Etterna watermark sound at the end.


tylerstraub:
Canceling this PR. Based on some feedback from Sarrrg we are going to revisit and add more emphasis on 16th notes.

PR #164, opened 2018-05-16
proposed sample accurate metronome based sync track (v3)

tylerstraub:
2nd PR request after feedback from Saarg. This version contains 8th notes on the click track.

> Based on a conversation in the EtternaDev Discord, we discovered that the existing file (_sync music.ogg) was inaccurate. The transients in the drum pattern were roughly 44ms behind. This music was created using ProTools and based around a sample accurate metronome. Also contains an Etterna watermark sound at the end.


tylerstraub:
canceling again. awaiting more feedback.

PR #168, opened 2018-06-13
Changed location of logged in text

11brendon:
changed the location of the logged in text at the bottom of the screen to above the session timer rather than to the right of it so it won't overlap


TheROPFather:
Nice, thanks.

PR #196, opened 2018-07-21
[Install] Make arrow shape uniform

foxfire667:
Sorry Rop, Joel caught some asymmetry with the arrow and made a new shape, Nick was holding off on the merge but you got to it already rip. 

Exact same layout, the arrow is just completely uniform and symmetrical now.


nico-abram:
:+1:

PR #199, opened 2018-07-24
Implementing Expandfield C++ into Etterna

Jousway:
adds coloured holds, stretch holds, shrink holds, overlay receptor (ovcepter), and some ryhtm based effects to receptor and explosion/ghost.


nico-abram:
Im not convinced about "ovceptor"
Merging anyways

PR #201, opened 2018-07-26
Removed 3:4 and 1:1 Aspect Ratios

poco0317:
There is no reason for them to exist in the game's current state. If you
select them then you get very broken results in the menus.

I also changed the order of the aspect ratios very slightly to make it pretty.
To quote Foxfire and Mina:
Foxfire | Today at 7:13 PM
3:4 and 1:1 must die
gratuitous double negative usage | Today at 7:13 PM
i would but who knows what will break


TheROPFather:
We discussed it and decided, that although it might not harm anything to remove them, they also cause no harm by existing. Might as well just keep them around for now.

PR #209, opened 2018-08-08
Make the Restart Gameplay Button Mappable

poco0317:
For your consideration: It has been suggested countless times that the Restart Gameplay button be able to be remapped. And it is able to be remapped in the game's current state. You can do it in the theme files. 
Doing that isn't good enough to me (and those who don't care enough to make that change).
I moved the key binding from the theme file to the actual game code, so it functions as a global button like the coin button, operator, etc.

The way to disable the button now would be to rebind the default key ( ` ) to something else like Coin instead of ripping out the theme code.

This change can be expanded upon possibly to make it function similarly to how you have to hold Enter for several seconds to exit gameplay.


poco0317:
Found a Lua error later on in WifeJudgementSpotting.lua due to these changes.
The originally commented out function (froot) is referenced again around line 679. This reference should be removed or commented.

PR #258, opened 2018-08-24
Somewhat Functional Replay Viewer

poco0317:
A popular requested feature I decided to try my hand at. I left the current "new" format of replays alone and utilized the highscore data they produce to tell autoplay to play files according to these replay files.
To do this, I reenabled PC_REPLAY, which appears to have not been enabled for .... ever?
I made it function half as an autoplay and half as a human player.
In using the vectors given by the existing highscore data from the replays, I drop the care for the hold section due to it seeming to be completely useless.

The button is here ingame in the score tab of a file with a compatible replay:
fasdfd

Once ingame, you are no longer allowed to use the restart button, and you cannot toggle autoplay using F8.

Known bugs:
1. Mines may fail the autoplayer if hit in dense mine fields
2. If multiple mines sit in one row (like a quad) and one is hit, the replay hits the left ones first
3. Roll notes count for a 0ms marvelous every "beat" allowing for > 100% plays.
4. Scores "persist" from these replay files but do not actually save (They disappear when you restart the game. Not sure if they attempt to upload).
5. Rare cases of crashing when charts are reloaded and you attempt to view the replay (This should be a current known bug)
6. Rare cases of crashing when charts are reloaded multiple times and you play the chart, then you view the replay, then you restart the game, then you try to view the replay again. (???????)
7. Turning autosync on or using other debug buttons probably breaks this
8. Rarely, the file is misread or offsync and notes are hit with 0ms accuracy in some sections.
9. Dropped holds and rolls don't actually drop. (The drops are not saved in the replay data)


poco0317:
wrong branch yikes

PR #309, opened 2018-10-04
Update 01 timing_windows_config.lua

Void345:
change the osu! marvelous rating to 3, as the 3.2 is only the weighting in score, not the accuracy


Void345:
it might also be nice to change the OD to 9 rather than 10 since 8.50-9 are more common

PR #310, opened 2018-10-04
Update metrics.ini

Void345:
change percentuntilcolorcombo to 0 because it looks better
like why would u wanna wait 25% thru to see if you have good combo


MinaciousGrace:
the point is to have a dividing point between two colors that can be changed

to set it to 0 is the equivalent is an opaque way of de facto removing the preference for anyone who doesn't know exactly where to look

an argument can be made for removing it and replacing it with an easily editable and transparent lua block but this isn't that

PR #313, opened 2018-10-04
preparation for country leaderboards

External69:
This is mostly front end; the changes to source code are just tiny fixes.

(the combo box (dropdown) doesn't do anything right now, and still needs to be moved to an appropriate place on the screen / redesigned)


MinaciousGrace:
committed directly with b7686d2794108c056df4e5f255316b628bafa02f and 57fd70e95826c55bebfd9c4d24848477d05c37f7

PR #362, opened 2018-11-20
Fixed critical namespace error that stopped from compiling

functional-lambda:
Well, the game didn't compile without this change. It seemed like something with the namespaces was messed up, and that was the case. It was just about changing TRUE/FALSE to `jpeg::TRUE` or `jpeg::FALSE`. 
After doing this, the game compiled perfectly and everything worked as intented.


nico-abram:
@functional-lambda : This doesnt compile (By CI) on both linux and windows

PR #396, opened 2018-12-01
Weak bandaid to stop noteskin changes from crashing in music select

poco0317:
Pull request made for the purpose of code review; not likely final.
All this does is stop the crashing you get if you change noteskins with an existing (visible or not) notefield. This will not change your noteskin unless you deliberately force the notefield to regenerate.


MinaciousGrace:
at this point its better to fix the source of the issue (hacky implementation of the feature) than bandaid all the peripheral issues it causes

Issue #449, opened 2019-03-01
(Linux, 0.65) Segmentation fault on startup since pull request #420 (because of libpng?)


kangalioo:
Etterna crashes on startup, with two warnings and an error from libpng. This happens with every build since commit 3311ff2 where pull request #420 was merged. Before that merge, there were no warnings or errors at all from libpng.

**To Reproduce**
Steps to reproduce the behavior on a Linux system:
1. Download 0.65 source code
2. Create "Build" directory and navigate into it
2. Build using `cmake .. -DCMAKE_BUILD_TYPE=Release -DWITH_FULL_RELEASE=1 && cmake .. && make -j6` (it's reproducible with `cmake .. && make` as well)
3. Execute the etterna executable

**Expected behavior**
I would expect that libpng doesn't throw an error, that there's no segmentation fault and that Etterna properly launches.

**Screenshots**
[Terminal output](https://pastebin.com/UX4qRbqu)
[Stack trace from gdb](https://pastebin.com/yhwU1JHZ)
[Terminal output from a working commit](https://pastebin.com/Vy4rCBFx)

**Desktop**
- x64 Arch Linux 4.20.11
- libpng 1.6.36-1
- glibc 2.28
- libavcodec: 0x372765 (3614565)

MinaciousGrace:
libpng has been replaced with an image library from the this decade but i guess not unlinking it was an oversight, doing so may fix the seg fault?

kangalioo:
How would I unlinking libpng? I've never done something like that

nico-abram:
Remove this folder: https://github.com/etternagame/etterna/tree/a4515afd3ef6c421067a381aa63231268f0ccb37/extern/libpng
These 3 lines: https://github.com/etternagame/etterna/blob/b8e69ffa4e083be265bf2501021cc9dbc90ea65b/src/CMakeLists.txt#L646
This line: https://github.com/etternagame/etterna/blob/b8e69ffa4e083be265bf2501021cc9dbc90ea65b/src/CMakeLists.txt#L332
These 2 línea: https://github.com/etternagame/etterna/blob/a4515afd3ef6c421067a381aa63231268f0ccb37/src/CMakeProject-texture.cmake#L45 and https://github.com/etternagame/etterna/blob/a4515afd3ef6c421067a381aa63231268f0ccb37/src/CMakeProject-texture.cmake#L54
[...]

PR #458, opened 2019-03-08
Linux Segfault Fix

jameskr97:
This is my second pull request on this repository.

1. Removed the link and source files for libpng exactly as described in #449


poco0317:
Before this is carried out, we need a replacement for the PNG saver/writer or need to make a decision what to do rather than use it (jpeg and bpm writers exist through libjpeg and rage).

PR #478, opened 2019-03-17
Fixed macOS includes, and updated macOS CI

jameskr97:
This is my fourth pull request on this repository.

Here is a summary of everything changed in this pull request.
- Fixed references in various OSX files which were not being included properly, disallowing the build to actually happen.
- Updated .travis.yml for macOS in order for it to have less verbose output, and also output as each file is compiled, rather outputting everything after the build is complete.
- Upgraded the `CMAKE_CXX_STANDARD` to 14. Even thought it is currently at 11 and building, the compiler or CMake is likely upgrading this for us, due to a specific line of code noted in the comment below.

Please let me know if there is anything I could do to help get this pull request merged.


StepmaniaCore.cmake
```diff
@@ -53,7 +53,7 @@ include("${SM_CMAKE_DIR}/SMDefs.cmake")
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
 # Enforce the highest C++ standard used in the code base
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 14)
```
nico-abram: I feel like this change is unrelated to this PR

jameskr97: While it isn't related to this PR, it is necessary to happen eventually, and I believe either CMake or the compiler is upgrading that for us. Note these lines of code:

https://github.com/etternagame/etterna/blob/6cbc98dc75f0805193e5d24e07ed01ef546bf049/src/NetworkSyncManager.cpp#L1112-L1113

m_PlayerReady is a `std::vector<bool>` and `emplace_back` was only available in `vector<bool>` in C++14. So whether we set it to 14 or not, it's gonna be upgraded to 14.

Note this Travis CI macOS log line: https://travis-ci.org/etternagame/etterna/jobs/507402245#L8712

It's one of our files being compiled, and if you go through it you'll eventually see `--std=gnu++14` If it wasn't obnoxiously verbose, I bet we'd see argument on every single line that compiles one of Etterna's source files.

I hope this explanation justifies why it's upgraded.

jameskr97:
This is ready to be merged in it's current state.

nico-abram:
Looks good to me

PR #489, opened 2019-03-23
Update RageSurface_Save_JPEG to use stbi_write_jpg

Bumrang:
Changed jpeg screenshots to use stbi_write_jpg from stb_image_write.h. 

Only place libjpeg was used, so I removed all references to libjpeg.

Compiled and tested on Windows 10, should work the same on other systems.


martensm:
LGTM, basically the same as how `libpng` was replaced.

PR #514, opened 2019-05-02
CMake Spring Cleaning

jameskr97:
This is my eighth PR on this repository.

Here is a summary of everything to be changed in this pull request.

# Required before merge
- Windows working with the following generators
  - [x] Visual Studio 15 2017
  - [x] Visual Studio 16 2019
  - [x] Ninja
- Static Analysers and formatters
  - [x] cppcheck
  - [x] clang-tidy
  - [ ] ~~clang-format~~ _(To be completed in isolated PR due to complexity of solving the issues this project has with consistent formatting, and not wanting to do a `Formatting changes` commit with a large diff)_
  - [ ] ~~include_what_you_use~~ _(Also to be completed in isolated PR as it requires significant changes)_
  - [x] link_what_you_use _(lwyu is a simple addition as it's a feature within CMake)_
- [x] Doxygen within CMake
- [x] LDoc within CMake
- [x] Resolve Windows error supression _(We get too many errors. We only want etterna specific errors.)_

# Other Fixes
- [x] libmad configured properly. Previously would stop certain packs from loading when attempting to delete references to audio files. (This specific bug occurred when attempting to delete a `RageSoundReader_FileReader` object at `Song.cpp:706`) [`15b505d`](https://github.com/etternagame/etterna/pull/514/commits/15b505dd872d6665a1716a336382b55d5fc6eed1)

PR #529, opened 2019-05-17
Fix MinaCalc linking logic

jameskr97:
MinaCalc_x86.lib would still attempt to link on macOS, even though it is a windows only lib. The compiler notices this, and choses not to link it, thought removing the possibility of it happening is better than letting the compiler deal with it.

Intended to be apart of #514.


nico-abram:
:+1:

PR #530, opened 2019-05-17
Eliminate foreach.h

jameskr97:
This PR should only me merged AFTER PR #514.

The foreach.h file contains several define statements to help shorten code when using for loops. Since the introduction of ranged for-loops, having such a file only increases the number of includes within the codebase. It did support using iterators across the entire codebase, but it was not necessary in more than half of the use cases. A ranged for-loop would have accomplished the same task, and in the few situations where an iterator is more suitable, it may be used where it is needed instead of writing a general use case in a different file, importing that file, and using the defined general case.

Here is a summary of everything to be changed in this pull request.
- The following define statements were removed
    - [x] FOREACH
    - [x] FOREACHD
    - [x] FOREACHS
    - [x] FOREACHM
    - [x] FOREACHMM
    - [x] FOREACHUM
    - [x] FOREACH_CONST
    - [x] FOREACHD_CONST
    - [x] FOREACHS_CONST
    - [x] FOREACHM_CONST
    - [x] FOREACHMM_CONST
    - [x] FOREACHUM_CONST

- [x] Every usage of the above defines have been removed.
- [x] Every `#include` references to foreach.h has been removed.


martensm:
Do you plan to address other `FOREACH` macros not defined in this file in this PR? Here is an example of one:
https://github.com/etternagame/etterna/blob/445898435e7edcf42aa407d26e43f091b4cdc36e/src/Etterna/Models/Misc/EnumHelper.h#L9-L14

jameskr97:
> 
> 
> Do you plan to address other `FOREACH` macros not defined in this file in this PR? Here is an example of one:
> 
> https://github.com/etternagame/etterna/blob/445898435e7edcf42aa407d26e43f091b4cdc36e/src/Etterna/Models/Misc/EnumHelper.h#L9-L14

I did not intent to remove that. I wanted to focus on removing a full file that was just helpers, and while that file is just enum helpers, to remove it completely would require more than than replacing statements with ranged for-loops. I'm not sure if there is a clean way to iterate through enum values with a ranged for-loop. I'll look into it and consider removing them, but at the moment, I do not plan to.

src/Etterna/Actor/Base/ActorUtil.cpp
```diff
@@ -459,9 +458,8 @@ ActorUtil::LoadAllCommandsFromName(Actor& actor,
 	set<RString> vsValueNames;
 	THEME->GetMetricsThatBeginWith(sMetricsGroup, sName, vsValueNames);
 
-	FOREACHS_CONST(RString, vsValueNames, v)
-	{
-		const RString& sv = *v;
+	for(auto const v : vsValueNames){
+		const RString& sv = v;
```
nico-abram: I think this should be a reference.

src/Etterna/Actor/Base/BitmapText.cpp
```diff
@@ -956,9 +955,8 @@ BitmapText::AddAttribute(size_t iPos, const Attribute& attr)
 	int iLines = 0;
 	size_t iAdjustedPos = iPos;
 
-	FOREACH_CONST(wstring, m_wTextLines, line)
-	{
-		size_t length = line->length();
+	for(auto const line : m_wTextLines){
```
nico-abram: I think this should be a reference.

nico-abram:
I think most of these should actually be using `auto&`. The ones which are ex. just a vector<T*> are probably fine without being a reference (Since any mutation on them would still affect the original objects, and since the copies are just as expensive afaik if not actually more expensive when making a reference which requires iirc an additional dereference(I'm talking about (T*)&) unless it gets optimized away)

nico-abram:
Btw compile on linux with gcc/clang seems to be failing due to the removed `#include "Etterna/Globals/global.h"` in the last commit (I think).

jameskr97:
Understood. I will open a new PR with similar changes, but pay more attention to whether or not the loop should be a for loop or a while loop. I'm not the best person to implement this type of change, since I don't have any experience with what datatypes/objects I'm changing and the best way for them to be accessed. I only can remove the file by trying to duplicate the way it seemed like it was accessed. Since there are likely many spots where I couldn't tell how it was done, I'll start over and pay more attention to exactly the best way for them to be accessed.

PR #532, opened 2019-05-20
CPack Hotfix - Install Directory

jameskr97:
This PR changes where the Etterna files are installed. Previously, the files were placed in
 `C:\Games\Etterna <version>\Etterna\<ALL_GAME_FILES>`. Since that middle `Etterna` directory was unnecessary, it has been removed.


martensm:
👍

Issue #380, opened 2018-11-26
Centralize asset customization into a single screen (toasty/noteskin/avatar etc.)


MinaciousGrace:

MinaciousGrace:
https://github.com/etternagame/etterna/pull/538 functional draft

PR #538, opened 2019-05-25
Add Asset Picker

poco0317:
This adds an Asset Picker to the game, simply. It allows not having to overwrite the default assets, basically. It's designed to work for any predefined Asset Folder in Assets Config.


MinaciousGrace:
related https://github.com/etternagame/etterna/issues/380

Issue #537, opened 2019-05-24
Changing devices (plugging or removing) causes game stutter


poco0317:
This is due to "Polling for Joystick Changes" usually, which lasts for 60 seconds.
Something should be done to either stop this or subdue the stuttering.

Relevant section of code (good old 10+ year old code): https://github.com/etternagame/etterna/blob/develop/src/arch/InputHandler/InputHandler_DirectInput.cpp#L841

Possibly related: #484

nico-abram:
Fixed in https://github.com/etternagame/etterna/pull/539

PR #539, opened 2019-05-25
Fix https://github.com/etternagame/etterna/issues/537

nico-abram:
No longer poll for a set amount of seconds/tries for device changes when using direct input (Windows only afaik), instead listen to the WM_DEVICECHANGE windows message and poll once when we get it.


poco0317:
Tested and works
(built x64 Windows 10, devices were a Vive and a usb xbox 360 controller)

PR #540, opened 2019-05-25
Update Building.md

TheKamilia:
removed a part of the apt command that would cause an error


poco0317:
The instructions edited here are replaced by newer instructions which no longer include libpng (because we don't use it)

martensm:
This is fixed on the `develop` branch, thank you anyways.

Issue #519, opened 2019-05-06
Add option to change Lane Cover color in game


LR2Zangoose:
Be able to change the color of the lane cover in game without having to adjust lua files and reloading the game

poco0317:
done a4a122a4aa762d5a3fd96a6f8f5d1d983c9d05e8

Issue #510, opened 2019-04-13
Add an option to enable/disable lifts when loading osu songs


Kangaroux:
The osu song loader adds a lift to the end of each hold. I think by default this should be OFF and the user has the option to turn it ON. The purpose of converting should be to play the song as you would in Etterna.

poco0317:
done 315cca5b071d58370354a982da6cf5b5a25360d9 4bfada032c155f514e8fd5e14c82dcd20b3221e6
added as an option to debug menu F8 as well 339b91b3b30403e1ca35d9a192e23e636d90f6fc

Issue #508, opened 2019-04-07
Fade notes in/out when using hidden/sudden


Kangaroux:
Holds fade in/out correctly but the taps themselves just instantly appear/disappear. It's a little jarring how the notes just flash on and off of the screen which kind of defeats the purpose of using hidden/sudden to reduce visual strain. Maybe this could be an option but I don't think it's intended.

Currently looks like **A**, should (in my opinion) look like **B**

![image](https://user-images.githubusercontent.com/2302541/55677682-305be180-58ba-11e9-9b43-e3e8259c8808.png)

poco0317:
fixed e12144ea68f5c147a38ba1accf6c63851cbb068c

Issue #464, opened 2019-03-11
Clean up the debug menus


poco0317:
Some debug options do not function or are not clear.
Add some more useful or entertaining options to the menus for testing and for actual usage purposes.

Manual screen rotation?
Certain advanced options toggles (pitch rates for one example)?
Fullscreen toggle?
Clearer volume modification?

Part of this will take place in Lua, but also in the Debug Overlay in the code.

poco0317:
dealt with somewhat 
efa0767a3a0912854264bc98f8050efc0725ce20
339b91b3b30403e1ca35d9a192e23e636d90f6fc

PR #541, opened 2019-05-25
Fix & Solve several Issues

poco0317:
Fix crash due to #460 
Allow osu lift toggling #510 
Add more useful features to debug #464 
Color config lane covers #519 
Fix abrupt fading on hidden/sudden #508


nico-abram:
Just skimmed this. Looks good to me :+1:

Kangaroux:
epic

PR #543, opened 2019-05-25
Remove `using namespace std`

jameskr97:
`using namespace std`, and any line starting with `using` that is meant to import a standard library object will be removed, with the exception of `using` statements for external libraries. Having direct references to what elements are and are not apart of the standard library.

# Required before merge
- Removing `using namespace std`
    - [x] Compiling on macOS
    - [x] Compiling on Linux
    - [x] Compiling on Windows


jameskr97:
The original purpose of this PR to remove all instances of `using <anything>` and remove all instances of `RString` was too broad, and it will be split up into multiple PR's.

PR #553, opened 2019-06-01
Upgrade to C++17

jameskr97:
C++17 brings new features to C++ that we might want to use in Etterna. It's a simple change, and only causes one in code change of removing `std::random_shuffle` to instead use `std::shuffle` as `std::random_shuffle` is removed in C++17.


MinaciousGrace:
self-reminder to merge after 0.69 release 

if i forget whoever sees this should yell at me

PR #567, opened 2019-06-06
Remove "using <anything>"

jameskr97:
Currently, every source file is reliant on the header file `Etterna/Globals/global.h`, and includes within that header file. Along with a globally included header file, `using namespace std` was added in that file, which allowed any function from the standard library to be used anywhere within the source with no direct connection to exactly where the function came from, which may be obvious for fields like `std::cout` and `std::endl`, though maybe turn out to allow for ambiguities with other included libraries.

This PR will attempt to remove all `using` statements, to move towards a larger goal of removing the `global.h` header altogether, as everything else within that header, can be modeled in a different way, and still accomplish the same task.


jameskr97:
I will close and redo this PR. I went off the old PR I had from before, merged develop onto this branch, then removed any remaining "std::", but it looks like some outdated code still made it through. I'll start over from latest develop, to make sure old code isn't also in other places in the source.

PR #569, opened 2019-06-09
ScreenEvaluation: Tweak Chord Cohesion text and add MA/PA ratio indicator

kangalioo:
This is my first pull request! I added an MA/PA ratio indicator in the evaluation screen and the Chord Cohesion message now only shows when Chord Cohesion is enabled (also, it is aligned to the rest of the text, contrary to how it was before)

That's what it looks like

Screenshot_20190610_001514


kangalioo:
I accidentally included the hotfix commits in the pull request... Mr. Nick said it might not be that big of an issue and I don't know how to fix it anyway... Should I create a new branch with only my commits, or is this fine?

PR #574, opened 2019-06-13
Fix ALSA-sw driver not loading

ThereGoesMySanity:
It looks like clang-format added some spacing that completely broke the ALSA driver.

I fixed the spacing and disabled clang-format on the offending line.


poco0317:
interesting
i did some digging and found the same issue present in directsound-sw
this is pretty funny, we didn't notice it for 9 months

PR #591, opened 2019-07-05
Experimentally Rewrite RageTimer Implementation

poco0317:
This is done with the hope of fixing some stutter, drift, or inconsistency issues.
More testing is needed, and I have no idea what I'm doing here. 

RageTimer hasn't been touched for many years, and it's looking a bit outdated (and with outdated comments). Some discussion may need to take place as for what to do with it exactly. What I've done to RageTimer here is change its internal time-keeping stuff with a std chrono duration to hopefully make things both more precise and more consistent. However, I don't know if that has actually done anything. All I do know is that it didn't break anything. (and for the things it did break, I fixed)

Some decisions I made in working on this (for example removing the bAccuracy parameter) are based on the fact that some of these comments and practices are pointing to things that are relevant for many years ago, and shouldn't be an issue in my opinion.

### DIFF
--- a/src/Etterna/Actor/Base/BitmapText.cpp
+++ b/src/Etterna/Actor/Base/BitmapText.cpp
@@ -356,7 +356,7 @@ BitmapText::BuildChars()
 	}
 
 	if (m_bUsingDistortion) {
-		int iSeed = lround(RageTimer::GetTimeSinceStartFast() * 500000.0f);
+		int iSeed = lround(RageTimer::GetTimeSinceStart() * 500000.0f);
 		RandomGen rnd(iSeed);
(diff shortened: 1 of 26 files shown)


poco0317:
Essentially ready for review. No notable performance improvements or resulting visible fixes.
Not sure if this fixes sync drifting over long periods of time. That's the last thing I haven't tested which I have been able to test.

nico-abram:
For anyone else reviewing this: I think everything above this (https://github.com/etternagame/etterna/pull/591/files#diff-f0fecd00f9c471e049460a811391bd9dR1) line is just formatting changes, removing an unused ragetimer from characters, and renaming function calls and is safe to ignore.

PR #594, opened 2019-07-11
Add more info on the Evaluation Screen

lotushomerun:
This edits the Etterna/Themes/Til Death/BGAnimations/ScreenEvaluation decorations/default.lua file:
Makes the difficulty, MSD, and SSR more clear by stating what they are
Adds  MA, PA, and CB ratios and will color coat them when you reach below (or above) a certain ratio

Example:
example

yes i know this is scuffed, only did it as a proof of concept

### DIFF
--- a/Themes/Til Death/BGAnimations/ScreenEvaluation decorations/default.lua
+++ b/Themes/Til Death/BGAnimations/ScreenEvaluation decorations/default.lua
@@ -534,136 +534,163 @@ function scoreBoard(pn, position)
 			}
 	end
 
-	t[#t + 1] = -- FA Ratio
+	if score:GetChordCohesion() == true then
+		t[#t + 1] =
 		LoadFont("Common Large") ..
-		{
-			InitCommand = function(self)
-				self:xy(frameX, frameY * 2.49):zoom(0.25):halign(0)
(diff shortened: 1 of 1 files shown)


poco0317:
I think that this would approach the line of adding too much information to the eval screen.

In addition, there doesn't appear to be rescoring or non-16:9 support in the changes made.

nico-abram:
I think the clarification test might be best as a kind of tool tip (If we can even manage to do that without too much work).

I'm not sure I see much point to MA or PA ratios since there's already %'s, but the CB ratio seems useful.

I don't like the hardcoded colours in this line onwards https://github.com/etternagame/etterna/pull/594/files#diff-f31a07d9b12f067c3d27a1a026279e02R571

nico-abram:
@lotushomerun Do you have any thoughts on our comments?

I don't think we want to merge this as-is for the reasons stated above. If you feel up to it, you could try to fix them, or if not, mention it so maybe someone else can work on it. Feel free to ask questions or discuss the proposed changes

poco0317:
Closed due to stale conversation. 

We don't want to add unnecessary information to the evaluation screen.

PR #597, opened 2019-07-18
Update Chinese Translation for Etterna 0.66.1 & Widely Enhance General Translation Support

SharpnelXu:
Planning on translating ColorConfig next.

Also formalized some translations to make everything consistent.

<details><summary>For those who are interested in what I did for consistency</summary>

不考慮翻譯 No Translations for these

- 	BPM
- 	NPS
- COMBO 連擊，但是擊的顯示有問題 (Can be translated but '擊' isn't displayed correctly)
- 	CPU
- 	StepMania
- 	遊戲模式類 Gamemode related
- 	成績類 Score related
- 	考虑不翻译MSD難度相關 Maybe MSD/difficulty related

Tabs

- General 信息
- MSD 譜面難度
-	Scores 分數
-	Search 搜索
-	Profile 個人資料
-	Filters 篩選
-	Goals 目標
-	Playlists 歌單
-	Packs 曲包
-	Tags 書籤

Mods

- 	modifier	特效
- 	attack		干擾，為了和特效區分開
- 	Receptor	判定區
- 	notefield	下落區
- 	Appearance	按鍵顯示
- 	Sudden		突現
- 	Hidden		消失
- 	Blind		盲打

	
General

-  Click 點擊->點此，因為擊有顯示問題
- 	Exit		離開
- 	offset		偏移値
- 	on			開
- 	off			關
- 	Course		組曲
- 	Group		曲包
- 	Autoplay	自動遊玩
- 	profile		個人資料
- 	UI			用戶界面
- 	option		選項
- 	server		服務器
- 	random		隨機
- 	preference	個人偏好選項
- 	package		整合包
- 	window		窗口
- 	fullscreen	全屏
- 	noteskin	按鍵樣式
- 	arrow/note		按鍵
- 	Combo Break	斷連
- 	Replay		回放

Sorts

- 	Preferred	喜愛
- 	Favorites	收藏

Scores

- 	PB			個人最佳
- 	High Score	最好成績

	
Difficulty, used old 3.95 standard

- 	Beginner	習
- 	Easy		樂
- 	Medium		踊
- 	Hard		激
- 	Challenge	鬼
- 	Edit		編

	
RadarValues

- 	Air			雙鍵
- 	Chaos		變化
- 	Freeze		長鍵
- 	Stream		流動
- 	Voltage		密度
- 	
- 	Taps		按鍵數
- 	Jumps		雙鍵數
- 	Hands		多鍵數
- 	Holds		長鍵數
- 	Mines		地雷數
- 	Rolls		連鍵數
- 	Lifts		離鍵數
- 	Fakes		假鍵數

MSD

- 	Overall			綜合
- 	Stream			單鍵
- 	Jumpstream		雙鍵
- 	Handstream		多鍵
- 	Stamina			耐力
- 	Jackspeed		速疊
- 	Chordjack		大疊
- 	Technical		技巧

FailTypes

- Immediate 立即結束
- 	ImmediateContinue	繼續遊玩
- 	FailAtEnd			歌曲結尾
- 	Off	關

Custom Gameply
ScreenGameplay

</details>


poco0317:
I went ahead and set up the ability to translate basically everything on SelectMusic and the Pack Downloader. I think it's about time the game could be fully translated. There's still a bit to do but that hits most of the bases.

Still to do are the abilities to translate the help (F12) page and the customize gameplay info.
Also need to add a translation string for the chat overlay.
Also need to add a translation string for the Select Profile title in the top left.
Also need to add translation strings for Asset Settings.
Also need to add a translation string for the Player Options title in the top left.
Also need to find a way to easily translate difficulties.

poco0317:
At this point all that needs to have translation support added that I am aware of is:
Color config
Tiny gameplay stuff
Customize Gameplay Explanations
Tip screen (F12)

poco0317:
All that remains is:
Color config
Difficulties (we may save this for the future)

SharpnelXu:
Asset Settings & FFWD/Rewind button still need translation support.
Will look into these tomorrow.

poco0317:
epic
As a note for testing which you may or may not have noticed: some strings do not reload when changing language unless you restart the game. The reason for this is because they are sitting in scripts. Particularly, the strings that follow this behavior are the skillset and radar value strings. If you switch languages while testing (though you shouldn't need to do it at this point) use Ctrl+F2 to reload scripts.
Everything looks good so far.

SharpnelXu:
Missing support for now (As to my knowledge):
Color Config

SharpnelXu:
Color Config Done.
Need to check for undisplayable characters and missed strings.

SharpnelXu:
Found some Strings in cpp files, will try to look into those later.

- **Name Playlist** when creating a New Playlist. [Link](https://github.com/etternagame/etterna/blob/425f10cfd58473754bba9021fc6f441e1accaefc/src/Etterna/Screen/Others/ScreenSelectMusic.cpp#L542).
- **Added chart** message header when adding a new chart to a playlist. [Link](https://github.com/etternagame/etterna/blob/425f10cfd58473754bba9021fc6f441e1accaefc/src/Etterna/Screen/Others/ScreenSelectMusic.cpp#L555).
- Multiplayer login screen. **Type Username**, **Login**. Also the option explanation is incorrect becuase it reuses the one for **Exit**. [Link](https://github.com/etternagame/etterna/blob/425f10cfd58473754bba9021fc6f441e1accaefc/src/Etterna/Screen/Network/ScreenSMOnlineLogin.cpp).
- **Username** and **Missing/Empty username or password** in the following prompt.
- **Create Room** and **Create a new room** in lobby. [Link](https://github.com/etternagame/etterna/blob/425f10cfd58473754bba9021fc6f441e1accaefc/src/Etterna/Models/Misc/RoomWheel.cpp).
- Tab names for Lobby.

poco0317:
Modifier translations may be tough to do. The string that shows here is the string that the game saves and parses to load your mods for every play. Localizing this could get messy. It might have to stay untranslated for a very-distant-in-the-future rewrite.

I've added translation support for the other stuff mentioned, though.
Also fixed the errant explanations on the login screen. It seems like the way we hid it is by just giving it a blank translation.
Also, I don't think it is a good idea to translate the "Missing/Empty username" string. At least for now. The reason for this is that this string is actually sent in the json response when you try to login to the etterna multi server. We should probably have it send a code instead and produce a string based on that code. Translating the string directly is probably not a good idea because that string could change at any given moment independent of the game.

SharpnelXu:
I thought we can just change the display, like spliting the modifiers then joining them together. Everything that should be translated is in OptionNames.

poco0317:
If splitting a string by commas is doable easily in Lua (and I'm 99% sure it is) then we can probably do it that way. We would just have to make sure that if no translation is found it doesn't return a blank. Like for the case of noteskins, there is basically no way to have those translated.

SharpnelXu:
I know I should add a new global function for mod translations somewhere...... Is there a Util lua class?

poco0317:
It might fit as a function in the fallback scripts files such as 00 utility or 02 utilities.

SharpnelXu:
Seems like note skins with special characters (space, comma, etc.) in their names will not be loaded correctly, so if that's not a bug then I don't need to fix that.

poco0317:
I think this is good for merge besides one possible thing that might be necessary to do.

The judgments' names are still hard coded in the fallback scripts `10 Texts.lua` file. These can be translated as well, if you would like. However it might be better to leave them like this so that we have a solid standard across the game. If translating them is not necessary, then this seems finished. Just needs to be tested a bit to make sure random little things don't cause errors.

SharpnelXu:
Let's leave the judgments out for now. We don't really have typical translations for them. I'd also like to modify color config translations tonight.

poco0317:
Yes, the color config translations can be changed to whatever you want in order to properly describe what each category and option does. It won't affect the actual game. 

I will wait for you to say it is ready to merge and then I will test things before merging.

SharpnelXu:
Done I think

poco0317:
Ready for merge after I redo a few of the translation things from the Lua side of things to keep loading the strings consistent and easy to mess with.
Performance impact with this pr seems relatively minor, definitely adds to Screen load time. The impact might be negligible after first load. If testers think it is bad, we will have to find a solution (load every string and distribute them in a more intelligent way than what is being done)

PR #600, opened 2019-07-21
Remove redundant license footers at the end of a ton of source code

External69:
someone had to do it


poco0317:
This is fine for merge but we may want to throw in a contributor file for Etterna-specific devs.

SneakySnake04:
Whether this is getting into the actual project or not....

This is a yikes. It violates the terms that come with the MIT License, which stepmania(and by extension, etterna) are in

'Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following coniditions: 

The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.'

Those license footers aren't redundant. They're required.

External69:
license text is still in the source code you don't need more than one

SneakySnake04:
> license text is still in the source code you don't need more than one

They're in multiple separate scripts and programs those of which are being used inside Etterna. They are needed.

poco0317:
The MIT license is distributed with the Etterna source. Adding the license to the specific source (when the whole source is under the license) is not a requirement.

poco0317:
Actually on second thought, it may be better to at least leave the names in their respective files with the copyright line and potentially remove the permission statement that goes along with each of them. The license is only required to be distributed with the source.

The point of this pull request is to remove the giant comments at the bottom of every source file. The supplementary credits file should be more specifically named and more specifically acknowledge the creators in some way.

Personally I would rather not remove them.

SneakySnake04:
There's already a file in the stepmania source that gives credit to the individuals involved.

And even then, this seems more like a personal gripe between you two that quite frankly nobody is complaining about. I see no real reason for this removal or addition outside of "I don't like it."

poco0317:
I have nothing to do with this, but this reaches too much of a legal grey area (for removing copyright notices) that I am willing to stop it.

We will revisit this later.

Poliwrath:
> The point of this pull request is to remove the giant comments at the bottom of every source file

maybe have "See LICENSE.MD for details"?

poco0317:
@SneakySnake04 it does exist in both the SM repo and ours, actually. We have just moved it.

https://github.com/etternagame/etterna/blob/cce9555823b0e2e5fdcaa92b632f42d27a41db94/Docs/legacy/credits.txt
https://github.com/stepmania/stepmania/blob/b95e49216eb2974b8e0f69b6603595df1c698ccd/Docs/credits.txt

nico-abram:
@SneakySnake04 There's no need to be insulting or have a negative attitude (`EDIT: even better: this is the only contribution you have made since signing up...hmmmmm`). It's not their first PR (Pherhaps in this account?), and even if it were, I'm not sure mentioning that is particularly relevant to the matter at hand.

I personally don't care too much about the license notices, I see *some* point to them since I think it makes it "crystal clear" what appllies to each file to anyone looking to use code in this repo (i.e until recently we had third party code under /src/, both added by us and inherited from the original repo, so it wasn't even very clear what was or wasnt third party code to which our license didn't apply. I *think* there is none at the moment, so everything under /src/ should fall under the given license. Regardless of whether we can or can't remove them, which I honestly don't know, I think an argument for per-file license notices can be made).

*However*, as a contributor mentioned a couple times, having files like https://github.com/etternagame/etterna/blob/develop/src/Etterna/Globals/Main.cpp#L10 where there's barely any code and the license notice takes up more lines than the actual code is kind of strange.

MinaciousGrace:
> I see no real reason for this removal or addition outside of "I don't like it."

this line would have a little more bite if it wasn't the operational logic behind everything the upstream devs do/don't do

you are right about one thing though, "this", is clearly about personal gripes

MinaciousGrace:
>I see no real reason for this removal or addition outside of "I don't like it."

yeah now it's about calling the bluffs of whiny old has-beens

Poliwrath:
image

TheROPFather:
Seems fine to me, good work.

TheROPFather:
Oops.

PR #604, opened 2019-07-26
Various tiny fixes and improvements (see within)

poco0317:
- SSM Lua had some updates that run too often so this helps that.
- SSM Lua involving Tags updated twice as often as it should so now it doesn't.
- Added the profiler to ms in sundries.
- Fix the game start bug by removing the ability for it to happen, and also hid some other stuff we don't use.
- Added GetTrueRotationZ for all Actors which we can use for button isOver later regardless of rotation.
- Removed the top 4 things from the key config because it confuses people
- Fix a Lua error that happens when you turn off Judgment text in Gameplay


nico-abram:
Seems good :+1:

poco0317:
Not ready for merge until it is confirmed that these changes do not break the multiplayer ready button functionality.
To reproduce: Go into Multi on this build. Press Ready. Have anyone else in the lobby press ready or pick a song. The color of the button may change to the wrong status.

poco0317:
Fixed button issue.
Undid sync overlay change because this is handled by the gameplay rewrite.

nico-abram:
This is blocked on https://github.com/etternagame/etterna/pull/597 to avoid conflicts on it

PR #605, opened 2019-07-28
Fix login message typo

Herkkupala:
add letter lmao nice.


poco0317:
This typo is handled in #597 which replaces essentially every string in the theme with a translatable one.

See: https://github.com/SharpnelXu/etterna/blob/82ec9455a257983aa6ba48e007b0117219ea9f4f/Themes/_fallback/Languages/en.ini#L196

Issue #601, opened 2019-07-22
Investigate mine behavior


poco0317:
Minebug.
Mines on different judges.

Is there something deeper at work? It seems that in some cases minedodge can be easier on a lower judge but also easier on a higher judge. Look into fixing minebug so that this is so inconsistent.

Minebug may happen because the past mine is being judged as a missed mine, when it should be judging the oncoming tap (providing you with a Miss)

poco0317:
Mine bug is fixed in gameplay rewrite by b3c8501e59ba2cdd59e1aa02c17e4b8ff94fc495
closed when merged

Issue #599, opened 2019-07-20
Investigate BMS gamemode behavior


poco0317:
There are reports that keysounds act oddly upon entering and exiting gameplay.

There are reports that customizegameplay does not function.
There are reports that gameplay itself has lua errors.

poco0317:
Mirror flips the scratch when it probably shouldnt?

Issue #595, opened 2019-07-18
Remove or Fix Blind Modifier


poco0317:
this is very broken and was effectively replaced by the Judgment Text toggle

intended behavior is that it just turns visual judgment feedback off but instead it causes notes to disappear instantly no matter what timing window you have (you can hit notes that arent on the screen)

poco0317:
handled in gameplay rewrite pr by 7d80865238aad48197dadc831b07087df420b7fd

closed when merged

Issue #593, opened 2019-07-10
Add different health systems


lotushomerun:
Different health systems would be a useful quality of life feature for the client.
# Proposing Health Systems
I propose 4 new health systems, batteries, non-regenerating health, hit-bar, and a accuracy threshold
## Batteries
Batteries work like their DDR/SM5 counterparts (You get a CB and you lose a battery. Batteries cannot be gained back in the same play, run out of batteries and you fail the file)
Normally, in SM5, you can only choose between 1-4 batteries
A slider like how you pick Cmod can solve this issue, by allowing you to pick from an unlimited range. You can have 1 battery, or 1,000,000 batteries
They're should also be multiple types of batteries:

- CB Batteries (CB = Lose a battery, like normal batteries)
- Perfect Batteries (Perfect or lower = Lose a battery)
- Great Batteries (Great or lower = Lose a battery)

These types of batteries will allow you to pick what score your going for (like are you going for a PFC?)

Oh and while we're at it, something like what issue #487 is proposing can be implemented too

## Non-Regenerating Health
This functions like Hard mode from LR2

- You start with 100% health
- CBs make you lose health (Affected by your Life setting)
- You cannot gain back health.

May be useful as an alternative to batteries

## Hit-Bar
A hit bar works like this

- You start with 0% 
- Hitting notes (NOT CBs) raises the bar, CBing drops the bar, but not as much as if you had a normal health bar (But is still affected by your Life setting).
- You must have 75% or higher bar to pass.

(Works like the way Sound Voltex does Effective Rate)

## Accuracy Threshold
An original idea that popped on my head:

- When you are below the accuracy threshold (Assuming that 5 or more notes have been played), a 5 second countdown will start. If you are below the accuracy threshold when the timer ends, you will fail the file
- You can select the accuracy threshold like how you can set the goal percentage in the goal tracker
- The accuracy threshold is for the judge your playing on. You can get a 92% on Justice and fail if your accuracy threshold is 93%, even though on J4, it would be a high wife AAA/low wife quad

# Conclusion
This might be doable on the theme, since it involves adding options to the theme and implementing them.
We are also bringing back SM5 features. That's cool, I guess

poco0317:
As a note, these features were previously in the game and were removed due to lack of use and feature bloat.

Related: #487

DeltaEpsilon7787:
[...]

PR #606, opened 2019-07-30
Refactor Gameplay, Replays, and Practice

poco0317:
Putting this up a little early to monitor progress and let people piss me off by commenting on my code before it's done

Handles #463 
(by extension handles #367 #461 #473 #474 #476 #491 #495 #496 #578 ~~#582~~ #583 #588 ~~#593~~ #595 ~~#599~~ #601 #616)
Handles #497 #444 #644


poco0317:
- [ ] In some cases, the Autoplay/Sync overlay does not update if the status is changed outside of gameplay.

poco0317:
As of this writing, replays are "fixed," at least with respect to the playback of them being scored incorrectly.
Time to move on to all the other issues.

poco0317:
As of writing, Chart Preview & Practice Mode sync is fixed. The drawback of this is that seeking is slower for long files. It was written in a way so that this only affects preview and practice seeking. This does not affect sample music playing.
The worst "lag" caused by this seems to be around 1-2 seconds for a 10 minute long seek. This is about the same amount of time it takes for sample music to start playing in normal situations.
An additional plus to seeking being written in this nonblocking way is that scrolling with preview opened will cause less stutter (due to loading audio and seeking far).

Kangaroux:
@poco0317 Is the bottleneck for seeking I/O related?

poco0317:
The changes to seeking are only relevant to MP3 files (and the other types are still capable of it, I just didn't change anything). What I did was force it to use an algorithm which seeks perfectly well forward at a reasonable speed, but if it has to go backwards the track must be started from 0 before seeking. The bottleneck here is perhaps in I/O but also how fast the CPU can decode the audio file either from 0 or from the point you are at depending on if you were trying to seek forward or backward.

See this commit and the relevant functions within SetPosition to see more info:
https://github.com/etternagame/etterna/pull/606/commits/7f6148d5a93ea181bfa8de42f6cdef3e869c992d

poco0317:
Currently working through new issues with Replay seeking and then Practice mode can be worked on.

poco0317:
Known issues that may not be fixed:
- Particularly unique files have sometimes unwatchable replays. One example is 1h54m JS, which uses an audio file that is literally empty. This breaks timestamp seeking or something and as a result, higher rates just don't work most of the time during replays. This is the only example I can find at the moment.
- The final WifePercent in a file may deviate from what it should be, but only by essentially 1 miss. This appears to be a rare bug possibly related to seeking. 
- Seeking in files causes a deviation in the wifescore based on the number of nontaps in the file. Particularly, seeking through files with many holds and mines even if they are dropped or not will cause this.

The expectation is that Replays work fine when watching from start to finish without seeking or pausing. So far that seems to hold true.

poco0317:
Anyways all that's left for Replays is:

- [ ] Restrict changing rates in a way so that Eval doesn't end up with wild numbers
- [ ] Verify that clicking the buttons doesn't click them 3 times (???)
- [ ] Handle setting mods temporarily for Replays; change mods back to what they were before entering Replay
- [ ] Secret feature
- [ ] Add a way to view the Replay from the Eval screen by clicking the graph (make this a choice screen to choose between practice probably)

Kangaroux:
@poco0317 for that third point, does that mean when you view a replay it uses the same mods as it was played with? Same scroll speed, perspective, mirror, hidden, etc.?

poco0317:
Yes. Mirror is actually already considered ever since one of the initial replay implementations. However, I would like to allow matching all mods except noteskin. The reason I didn't in the past is because of the complicated nature of exiting Gameplay/Evaluation in multiple ways. I'm more willing now to force it to work.

At the moment, only rate and mirror is recognized. 

Complications arise when trying to consider additive, random, or removing mods. As long as the mods work consistently, it should be fine.

Kangaroux:
My only concern is if you exit the game mid replay, will it overwrite your mods with the one from the replay?

poco0317:
Yeah the point of being thorough is to make sure that exactly what you said doesn't happen.

I'm taking some necessary steps which should make it clearer whether the player is backing out of gameplay (service switch or otherwise), quitting/ending to eval, restarting gameplay, or any other state I might have left out. From there, handling the cases should be simple. This may end up requiring introduction of a new ScreenEvaluationReplay. My expectation is not having to do that.

Kangaroux:
Instead of trying to cover every use case (which will ultimately fail if you kill the process since you can't react to that), can you leave the player's mods untouched and instead change what mods the game looks at when entering a song?

poco0317:
Killing the process does not save the player mods, so it wouldn't matter. Mods save with the XML, which typically only saves when exiting Evaluation.

It is possible to change just the gameplay mods.

poco0317:
As of writing, Replays are capable of generating their own LifeGraph and ComboGraph. They also apply any mods used during the original score. This specifically can be toggled off in Advanced Options.

Needs testing with mirror and permamirror.

poco0317:
As of writing, combo and life graph recalculation should be a bit more bug free. Still not perfect. 
Scores that use mods which make it difficult to calculate misses (anything that isn't mirror or like nomines) will not generate a graph. Also, CCon scores will not recalculate a graph. This is on purpose.

Starting to fix other tiny bugs and make enhancements now.
Moving on to Practice very soon.

poco0317:
Worked a bit on Practice instead of finishing the cool replay stuff. As of writing, Practice Mode now has a working loop region instead of strict bookmark. Setting the loop region can be done by either pressing Insert Coin to mark the current position, right clicking 2 spots on the graph, or control left clicking 2 spots on the graph. Precise marking can be done by right clicking outside the graph to pause and using the scroll wheel to go through the song at your own leisure, then press Insert Coin to mark the current position.

Loop regions are set by waiting for a first position. If a single position is given, it acts like a bookmark: You can press backspace to jump to it and play until the file ends. Rinse and repeat. If a second position is given using any method, the loop region is created from those 2 end points. Notedata is now restricted to this space only, and the song will loop endlessly in that region (as well as grant a 2 second intro and a 5 second outro so you can see your score). Stats reset every loop and every time you seek. You can press Backspace at any time to immediately jump to either your bookmark (shown as a blue line) or to jump to the start of the loop region (shown as a purple box).

Reset the loop region to bookmark-only form by double right clicking the graph, double ctrl clicking the graph, or double tapping Insert Coin.

I plan on allowing exiting to evaluation soon. Every seeking desync issue specific to practice mode should now be eliminated.

Kangaroux:
@poco0317 I noticed an issue in practice mode where even though the judgement counts would reset, it would seem the total note count used to calculate the % wouldn't reset. Was that also fixed?

Also thank you for the looping that is going to be a big help

poco0317:
All stats truly means all stats. A couple of commits have now been pushed which should make sure the visual side of things updates properly to reflect the reset stats.

poco0317:
Calling this effectively done and ready for a release.
Did not finish #582 #593 #599
Will handle later during development.

PR #611, opened 2019-07-31
Replace Ragelog internals with spdlog

poco0317:
Done for simpler internal usage and probably more efficient/faster stuff I dunno I just wanted something to do

(also probably fixes the Stack Overflow crash)


poco0317:
Current issues with this PR as of writing:

- [ ] Logging from other threads (?) seems to not work. This means that logging from SM itself, such as from ScreenGameplay or so on works fine, appears in stdout console, and the log file and flushes. Logged checkpoints, fps logs from the display (thread?) and maybe some other smaller things do not appear anywhere in the log. Breakpoints indicate that the code reaches the `LOG->Trace(` but nothing happens.

- [ ] Doesn't seem to build on appveyor/travis for various reasons related to the convoluted nature of Ragelog/RageUtil itself. May require a full rewrite and reimplementation of the class. It does compile locally, which confuses me further.

nico-abram:
Probably irrelevant, but going over this, I noticed I think https://github.com/etternagame/etterna/pull/611/commits/80ce8cbbaaf2d4c3371107ac5b7fa1bb759afa77#diff-d2c67078a65380ecfa471c1860005491R514 is making an unnecessary copy of the string (The overload that takes the &string is only using format.c_str() so if we write the &string one in terms of the char* one calling c_str() we can probably avoid a copy). But I'd be surprised if it matters at all

nico-abram:
On travis failing:
It's failing on `use of undeclared identifier '_isnan' (format.h:323:12)`. It seems this line is commented https://github.com/etternagame/etterna/pull/611/files#diff-0438d190cedefa196df8d1db57eb6fc0R279 but its uncommented in the spdlog master branch (https://github.com/gabime/spdlog/blob/master/include/spdlog/fmt/bundled/format.h#L433). Maybe we could try just upating the spdlog checked in to the repo? Just a guess

poco0317:
2 months later

spdlog fails to build on windows without the line in question commented
https://github.com/etternagame/etterna/pull/611/commits/f904fd985ea74669e7e5d5b3db6a2eb83da22e2b

jameskr97:
Maybe this should be closed? spdlog is C++11, and we're on C++14 moving to C++17 soon.

poco0317:
abandoned in lieu of #767

PR #617, opened 2019-08-03
Remove Characters: DancingCharacters, Character, and CharacterManager

poco0317:
I woke up today and something in my body was just telling me "clear your mind of the code you do not understand" so I decided to yeet this code

(characters are old, taking up space, and probably have no use in etterna at this point.)


nico-abram:
This is blocked/waiting on https://github.com/etternagame/etterna/pull/591 and https://github.com/etternagame/etterna/pull/606 (Since it's pretty low priority and it would introduce conflicts for them)

PR #640, opened 2019-10-04
Windows CrashHandler: Create a thread for separate stack

Prcuvu:
On Windows, every thread has a separate stack space. It is possible to create a separate thread to do the exception handling job without having to mess with the stack pointer. Therefore we can happily remove any inline assembly.

Special thanks to [sigatrev](https://github.com/sigatrev) for inspiration.

### DIFF
--- a/src/archutils/Win32/Crash.cpp
+++ b/src/archutils/Win32/Crash.cpp
@@ -294,8 +294,8 @@ RunChild()
 	}
 }
 
-static long
-MainExceptionHandler(EXCEPTION_POINTERS* pExc)
+static DWORD WINAPI
+MainExceptionHandler(LPVOID lpParameter)
 {
 	// Flush the log so it isn't cut off at the end.
 	/* 1. We can't do regular file access in the crash handler.
@@ -313,6 +313,7 @@ MainExceptionHandler(EXCEPTION_POINTERS* pExc)
 	 * However, once in a while some driver or library turns evil and unmasks an
 	 * exception flag on us. If this happens, re-mask it and continue execution.
 	 */
+	PEXCEPTION_POINTERS pExc = reinterpret_cast<PEXCEPTION_POINTERS>(lpParameter);
 	switch (pExc->ExceptionRecord->ExceptionCode) {
 		case EXCEPTION_FLT_INVALID_OPERATION:
 		case EXCEPTION_FLT_DENORMAL_OPERAND:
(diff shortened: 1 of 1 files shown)


nico-abram:
This seems good to me.

My only minor nitpick is, since we're handling the (However unlikely) scenario of not being able to start the thread (I would guess because there's no memory for the stack? No idea), shouldn't we also output something mentioning *why* we're terminating (A message box, console msg, or log entry)? I'm not actually sure if attempting that with the current stack is a good idea, and just terminating ASAP in this case seems fine to me anyways because of the stuff mentioned in the comment.

This seems strictly better than what we had anyways since we were ignoring the stack issues in x86-64. I do wonder how spinning a new thread interacts with sigatrev's concern for OOM exception handling (Won't that *also* fail to allocate the stack for the thread? I honestly have no idea. It's not like there's much we can do anyways)

I'm guessing this is a backport (Or forward port? No idea what the term is since we're the fork :P) of your PR in the sm repo. I just wanna thank you for it :yum: 

cc @poco0317 in case you're interested in looking at this before merging (Also, would this warrant a changelog entry? It's not something a user would ever be concerned with but I'm unsure of what the criteria is)

Prcuvu:
According to sigatrev, we should have enough remaining stack memory to launch a thread. Perhaps even enough for crash handling. So if `CreateThread` also fails, I seriously do not think there will be anything else we can do. Just let it crash.

This is indeed a port from my PR in StepMania repo, because some Etterna players I know recently found that PR and asked me to do the port. Maybe you should go thank them :)

poco0317:
Every change warrants a changelog entry. Since one doesn't exist for .67 yet I'll just have to add it later when reviewing everything.

Will merge later assuming everything goes smoothly (I was the one who saw the PR and mentioned to the guy who probably talked to you)

Prcuvu:
I have just finished revising VDI format for x64 symbol lookup on StepMania repo. Concerning that you have already moved to using ILK, do you expect me to port it here?

MinaciousGrace:
hello im alive

nico-abram:
I've honestly never looked at SM's VDI format thing, but any fixes/improvements are always welcome. Though maybe a separate PR would be a good idea, if you do decide to do so. Do you mind if we go ahead and merge this one as-is? (Probably after some testing just because)

Prcuvu:
I am fine with merging this one.

Issue #642, opened 2019-10-13
X11 Polling Issues


192314:
I've built the most recent version (0.66.1) for Linux, and when I play there are weird diagonal lines on the timing graph. It looks like the input is only being polled every 16.6 ms (the same as my display refresh rate).

According to [this](https://github.com/geefr/stepmania-linux-goodies/wiki/So-You-Think-You-Have-Polling-Issues) guide, [these](https://github.com/stepmania/stepmania/commit/9bd2770fed48c444c31dd578fb3ff10229423d93) changes are needed to have the polling rate not be limited by the framerate for X11 input.
(It looks like these changes preserve the event timestamps when passing the keyboard events to the game.)

PR #643, opened 2019-10-14
Preserve event timestamps in X11 input driver

192314:
Fix for #642 .

Analogous to [this](https://github.com/stepmania/stepmania/pull/1702) pull request:

> The X11 XKeyEvent structure (and others) include a timestamp field, with 1ms resolution.
> This was previously ignored causing the input timestamps to be set to the middle of each frame.
> If the timestamps are passed through then the event timestamps are not coupled to the framerate and the polling issues are resolved.

### DIFF
--- a/src/arch/InputHandler/InputHandler_X11.cpp
+++ b/src/arch/InputHandler/InputHandler_X11.cpp
@@ -344,7 +344,7 @@ InputHandler_X11::Update()
 				break;
 		}
 		if (real_release) {
-			ButtonPressed(DeviceInput(release_dv, release_db, 0));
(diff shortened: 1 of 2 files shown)


poco0317:
Looks fine. This doesn't concern me particularly, but I just want to point out that the commit listed isn't under your name.

Don't really have a way to test this other than confirming build is fine and having testers check it out.

nico-abram:
Seems good to me too (Just from skimming it, no testing either)

Should crediting be done in the commit message to the PR you linked in the message here?

192314:
> Should crediting also be done in the commit message to the PR you linked in the message here?

Sure, I just updated the commit message.

PR #652, opened 2019-11-25
Re-add Git Hash & Version Info Variables

poco0317:
At some point during the CMake rewrite for 0.65, these disappeared and were not replaced. This resulted in blanks showing up all over the place and especially in crash logs.
Now, it should show the version (no extra work is needed, it is pulled from the project version) and the exact commit/tag the build is from. Date and Time are removed.

Found a little typo that didn't affect anything, and I fixed that too.


poco0317:
Tested and works (version number and commit/tag shows up in crash info)

Issue #655, opened 2019-11-30
X11 input broken


zedeus:
After #643 input is partly broken. Inputs mostly work fine in the menus, like up/down etc, but some double key inputs don't (eg. double up/down to change difficulty), while others do (opening profile options with double enter). When playing a chart inputs are visibly registered but it always misses. Removing the `timer` argument from this line fixes it, but that essentially undos the PR.
https://github.com/etternagame/etterna/blob/4cf772eb9753c0adaf0c2086c85aacb618a867d3/src/arch/InputHandler/InputHandler_X11.cpp#L434

I guess it might be worth noting that this if-statement is always false with the timer argument.
https://github.com/etternagame/etterna/blob/4cf772eb9753c0adaf0c2086c85aacb618a867d3/src/arch/InputHandler/InputHandler.cpp#L18-L25

poco0317:
The if statement in question is meant to kind of produce a timestamp half way between the current update and the previous if no timestamp is given (hence the whole "timestamps bound to framerate" thing). If a timestamp is given, then that doesn't need to happen. That kind of breaks the behavior of `m_iInputsSinceLastUpdate` I guess. Maybe.

Missing in gameplay despite visibly pressing the arrows is very strange to me. The only thing I can think of that would cause that is `INPUTFILTER->ButtonPressed( ... )` coming back true yet having no button press. (to test this, you may not be able to press any single taps but you would still be able to finish a Hold or hit a Mine)

I'm interested in debugging this by knowing what the output of `event.xkey.time` is. It could somehow be possible that the way the timestamp is constructed here is wrong. https://github.com/etternagame/etterna/blob/4cf772eb9753c0adaf0c2086c85aacb618a867d3/src/arch/InputHandler/InputHandler_X11.cpp#L431-L432

In Windows' case of DirectInput, these timestamps are given and that block of code is never entered.

I can't test anything myself outside of Windows.

zedeus:
Mines work but Holds don't.

`event.xkey.time` is in milliseconds like the comment says, here are some examples:
```
2713654334
2713654417
2713654676
2713654755
2713655250
2713655314
2713655333
2713655385
```

I'm not 100% sure and it's probably not relevant, but I think these timestamps are relative to when the X11 server was started.

poco0317:
[...]

PR #656, opened 2019-12-01
Revert "Preserve event timestamps in X11 input driver"

poco0317:
Reverts etternagame/etterna#643

Pending conclusion of: #655 

If merged, reopen #642


zedeus:
Can you do a hotfix release with this included? It breaks input on multiple machines I've tried running different Linux distros

poco0317:
Yeah if more than 1 person could confirm that it causes this issue.

zedeus:
This can be closed

PR #659, opened 2019-12-04
Fix git hash missing from CI builds

poco0317:


poco0317:
tested locally and works
tested from 64bit appveyor works

PR #686, opened 2020-01-10
Fix macOS Catalina Crashing

jameskr97:
macOS Catalina had a change in their OpenGL API. Specifically one which crashes game if mac related UI updates are done off of the main thread. The issue is fixed by running the line which crashes the application in an asynchronous block to be sent to the main thread for execution.


poco0317:
I don't see anything particularly wrong with it, and it seems like the function used is supported on 10.6+

If it runs for you it's probably good. We might want to get someone on something older than Catalina to test, but that would probably be a hassle.

jameskr97:
> I don't see anything particularly wrong with it, and it seems like the function used is supported on 10.6+
> 
> If it runs for you it's probably good. We might want to get someone on something older than Catalina to test, but that would probably be a hassle.

The `dispatch_async` and `dispatch_get_main_queue()` functions are a part of Apple's Dispatch framework, previously known as Grand Central Dispatch. It's been supported since 2009 in one form or another. But definitely if other's running Catalina could also test, that would be very appreciated.

lmiller1990:
I am struggling to even build for Catalina. I made an issue about it - #718. How did you get it to run on Catalina?

jameskr97:
> I am struggling to even build for Catalina. I made an issue about it - #718. How did you get it to run on Catalina?

I was building with xcode. The issue you are experiencing in #718 is a problem that happens when linking with ninja specifically.

nico-abram:
@jameskr97 Do you think we should merge this before the next release since it makes the game crash on latest macos? This has not been tested on older versions yet, right?

Aside, could you explain the change to KeyCodeToChar? Did you try using different keyboard layouts/languages in case something broke?

PR #728, opened 2020-04-05
Curl Update to 7.69.1

jameskr97:
Switch curl to build from source, and updated to the lastest version.


nico-abram:
@jameskr97 do you know why the mac travis build failed? I retried it in case it was a false positive but it failed again. The error seems completely unrelated to the change, could it be related to some kind of preprocessor define conflict?

PR #736, opened 2020-04-18
Add SoloCalc

handbeezy:
added ix's SoloCalc
updated CMakeLists for SoloCalc
wifeTwirl and Steps updated to allow MSD display for solo StepsType
did clang format/tidy on sources and headers

### DIFF
--- a/src/Etterna/Models/StepsAndStyles/Steps.cpp
+++ b/src/Etterna/Models/StepsAndStyles/Steps.cpp
@@ -420,6 +420,10 @@ Steps::CalcEtternaMetadata()
 float
 Steps::DoATestThing(float ev, Skillset ss)
 {
+	// This is 4k only
+	if (m_StepsType != StepsType_dance_single)
+		return 0.f;
+
 	Decompress();
 	const vector<int>& nerv = m_pNoteData->BuildAndGetNerv();
 	const vector<float>& etaner = GetTimingData()->BuildAndGetEtaner(nerv);
@@ -450,6 +454,11 @@ Steps::GetCalcDebugOutput()
 	//	return;
 	calcdebugoutput.clear();
 	// function is responsible for producing debug output
+
+	// This is 4k only
+	if (m_StepsType != StepsType_dance_single)
+		return;
+
 	Decompress();
 	const vector<NoteInfo>& cereal =
 	  m_pNoteData->SerializeNoteData2(GetTimingData());
@@ -1007,7 +1016,9 @@ class LunaSteps : public Luna<Steps>
 			  L, CalcPatternModToString(static_cast<CalcPatternMod>(i)));
 			lua_createtable(L, 0, 2);
 			for (int j = 0; j < 2; ++j) {
-				vector<float> poop = p->calcdebugoutput[j][0][i];
+				vector<float> poop;
+				if (!p->calcdebugoutput.empty()) // empty for non 4k
+					poop = p->calcdebugoutput[j][0][i];
 				LuaHelpers::CreateTableFromArray(poop, L);
 				lua_rawseti(L, -2, j + 1);
 			}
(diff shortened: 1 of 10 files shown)


src/Etterna/Models/StepsAndStyles/Steps.h
```diff
@@ -10,6 +10,7 @@
 #include "RageUtil/Utils/RageUtil_CachedObject.h"
 #include "Etterna/Models/Misc/TimingData.h"
 #include "Etterna/Globals/MinaCalc.h"
+#include "Etterna/Globals/SoloCalc.h"
```
nico-abram, line 13: Is this include necessary?

src/Etterna/Models/StepsAndStyles/Steps.cpp
```diff
@@ -421,7 +425,8 @@ Steps::BorpNDorf(int modType)
 	dumbthings.clear();
 
 	if (modType < CalcPatternMod::ModCount && modType >= 0)
-		MinaSDCalcDebug(cereal, GAMESTATE->m_SongOptions.GetSong().m_fMusicRate,
```
nico-abram, line 424: I'm not 100% clear on what this call does but should this not have a solo equivalent? I could be missunderstanding something

src/Etterna/Models/StepsAndStyles/Steps.cpp
```diff
@@ -851,10 +856,14 @@ class LunaSteps : public Luna<Steps>
 		auto& ni = nd.SerializeNoteData(etaner);
 		if (ni.size() == 0)
 			return 0;
+		std::vector<float> d;
 
-		std::vector<float> d = MinaSDCalc(ni, rate, goal);
-		auto ssrs = d;
+		if (p->m_StepsType == StepsType_dance_solo) // Single SSR at the moment
```
nico-abram, line 861: Could you ellaborate on this comment?

nico-abram:
Hey @handbeezy 

I don't know that we necessarily want to merge a solo calc, however I thank you for the work you put in this. Even if we do, I don't know to which standard we would want to hold such a thing (i.e, would we be fine with merging a solo calc that hasnt seen that much work/testing (I don't actually know how much work has gone into it as-is)), and I don't know how this current integration interacts with EO at the moment (It should not interact with EO at all before it gets merged, I think)

On to the PR itself, I'm gonna ignore the actual calc contents and just look at the integration (I think even if we don't merge it into the main development branch we might want to have a branch with this as a testing bed for messing with a solo calc, I'm not sure though).

src/Etterna/Globals/NoteDataStructures.h
```diff
@@ -0,0 +1,39 @@
+// Only NoteData structures independent of the rest of Stepmania live here
```
poco0317, line 1: This entire file already exists at `/src/Etterna/Models/NoteData/NoteDataStructures.h`

src/Etterna/Globals/SoloCalc.h
```diff
@@ -0,0 +1,22 @@
+//
+// Created by Robert on 11/20/2019.
+// Edited by five 04/2020
+
+#ifndef MINACALC_SOLOCALC_H
+#define MINACALC_SOLOCALC_H
+
+#include "NoteDataStructures.h"
+
+typedef std::vector<float> SDiffs;
```
poco0317, line 10: These definitions already exist in MinaCalc.h

src/Etterna/Models/StepsAndStyles/Steps.cpp
poco0317: This was a formatting change. There needs to be a Solo equivalent.

src/Etterna/Models/StepsAndStyles/Steps.cpp
handbeezy: That probably shouldn't be there, the SoloCalc() call below this line used to only return a float for a single rating, not a vector with different ratings for all skillsets. I ended up just duplicating that one rating for all skillsets because I assumed something would mess up down the line.

nico-abram:
@handbeezy Do you think you'll have time to address comments around https://github.com/etternagame/etterna/pull/736/files#diff-6008f660ef81a40a6180a5a40d71765bL424 and https://github.com/etternagame/etterna/pull/736/files#diff-bcbec2a55edffb0fbdb01e29055474ddR13 soon? Aside, have you tested how this interacts with the debug calc output and EO uploads?

src/Etterna/Models/StepsAndStyles/Steps.h
poco0317: Yes

poco0317:
I would like to merge this to the calc development branch, so that we can release both this and that at the same time.
The changes I've made should bring it up to date with what it needs to work at a base level.

PR #744, opened 2020-04-19
Remove the for_mingw extern folder

poco0317:
what does this do


jameskr97:
It does nothing. Please delete.

PR #756, opened 2020-05-07
Wife 3 release prep

MinaciousGrace:
test version that sands off the rough edges in preparation for release version
everything is rerouted to old calc


MinaciousGrace:
almost everything is fixed and ready with the exception of display mismatches on eval with internal score due to missing mine hits and hold drops

PR #758, opened 2020-05-13
Change files in VS, NetworkSyncManager to use /bigobj

jameskr97:
The VS solution explorer was updated to have as close as possible structure to the filesystem for where the files are located. FIles which windows does not compile are not shown.

NetworkSyncManager is a large file and will not compile in debug mode without defining it as "/bigobj." That compile flag is set in `CMakeWindows.cmake`.


poco0317:
Using Visual Studio 2017, emptying Program and build folders for each cmake rebuild and recompile:
Release compiles on x64.
Release and Debug fail to compile on x86 due to unrelated reasons (unresolved externals from MovieTexture_FFMpeg.obj)

Debug fails to compile on x64 due to missing `/bigobj` for NetworkSyncManager.cpp (what this change should have taken care of)

poco0317:
Debug still fails after the changes. RelWithDebug makes a return and still works.

jameskr97:
Better version of this incoming.

PR #768, opened 2020-05-20
Improved CMake System

jameskr97:
# Better CMake

Last year when I rewrote CMake, it was an improvement, but there weree somet things pushed off until later. This PR is the "later."

## Changes Include
- Minimum CMake version upgraded to 3.15.0. This gives us access to cmake policy [CMP0092](https://cmake.org/cmake/help/latest/policy/CMP0092.html), which removes default compile warnings from extern projects.
- All extern library variables were moved to extern CMakeLists.txt, as that is a more suitable location for those variables. They are paired with the `add_sudirectory` of which they affect.
- The C run-time is statically linked to the game. While not recommended, when considering that most installations of Etterna are portable, having the CRT be portable as well could cause less issues for end-users.
- Visual Studio filters actually match the file system structure. (It has to be organized after the source files are added to the executable
- All libraries are explicitly declared private instead of being implied as was done previously.
- Lua targets were separated int a different IDE folder. Only for organization.
- BUILD_SHARED_LIBS set to off explicitly as opposed to being implied before. Libraries built from source should no longer try to make a `dll` file.
- Statically link OpenSSL.
- All 4 build configurations (Release, Debug, RelWithDebInfo, and MinSizeRel) will build and run.
- Removed large section checking `CMAKE_BUILD_TYPE` as it is not correct to be used with a multi-configuration generator such as CMake, It has been replaced with generator expressions. (A large portion is no longer necessary as CMAKE_MSVC_RUNTIME_LIBRARY is given a default value for static linking in the room CMakeLists.txt
- Remove use of CMAKE_GENERATOR_PLATFORM to make project generation easier.
- Defining `OPENSSL_DOOT_DIR` is no longer necessary on Windows (assuming you read the instructions, and install the correct version of OpenSSL that CMake nativley supports)
- Build instructions have been updated to reflect the new changes.

## Before merging
- Make sure #780 is merged. That PR has already been merged into this one.


poco0317:
In its current state using Windows 10, VS2017:
CMake required an update, normal. Not including the openSSL directory worked fine.
Debug, MinSizeRel, Release, and RelWithDebInfo x64 fail due to unresolved extern pointing to libcrypto.
Exactly the same on Win32/x86

... appveyor fails for the same reason

poco0317:
All targets "work" in vs2017/2019 x86/x64, but only after updating openSSL (to 1.1.1g)
It appears that the documentation has been updated to match the version change. May need to clarify that it is required if not already.

In x64 and win32 Debug, consistently crashing when closing the game. Applying  612ee7f51e173b120d45a302444a6facc66582a3 fixes it
The win32 builds run after some messing around with the Program folder. A clean clone and run should be fine.

PR #777, opened 2020-05-25
Replace fftw with Mufft

nico-abram:
Mufft has a more permisive license, and is also dramatically simpler than fftw

I did some basic checking of FPS on selectmusic and didnt notice big differences


poco0317:
its faster

nico-abram:
Moved to https://github.com/etternagame/etterna/pull/790

PR #783, opened 2020-05-28
Convert fully to Github Actions

jameskr97:
# Goodbye, Appveyor

After some more tinkering, specifically learning more about what chocolatey can do, a 32bit build is now possible on Github Actions (by forcing choco to install 32bit OpenSSL). We initially used OpenSSL libraries included within Github Actions, but is not the same as what we were compiling with. Chocolatey has the correct version, and after installing, proper static links to OpenSSL were attached to the binary.

## Notes
- Since Appveyor is still activated on this repository, appveyor will still pull and attempt to build this PR. Please ignore the results of that CI. It will likley fail.

## Changes include:
- OpenSSL static linking correctly.
- 32Bit build and artifact on Github Actions
- `.appveyor.yml` removed.


.github/workflows/ci-workflow.yml
```diff
@@ -85,22 +81,27 @@ jobs:
...
+        if: ${{ matrix.cfg.arch == 'Win32' }}
+        run: choco install ninja nsis openssl -y --x86
+
+      - name: Install chocolatey packages (x64)
+        if: ${{ matrix.cfg.arch == 'x64' }}
+        run: choco install ninja nsis openssl -y
 
       - name: Generate CMake
-        run: mkdir main/build && cd main/build && cmake -G "Visual Studio 16 2019" -A "${{matrix.arch}}" ..
+        run: mkdir main/build && cd main/build && cmake -G "Visual Studio 16 2019" -A "${{matrix.cfg.arch}}" -DOPENSSL_ROOT_DIR="${{matrix.cfg.ssl-dir}}" ..
```
nico-abram, line 93: Is OPENSSL_ROOT_DIR needed again? Or is this just needed in CI? If it's needed again, we should update the build instructions right?

jameskr97: It’s just for CI, since CI has multiple 64Bit OpenSSL versions. CMake uses OPENSSL_ROOT_DIR as a hint to know where to look for it. If you one have one OpenSSL that is in the “cmake OpenSSL search path” you don’t need it. But we have two, so it’s used to distinguish.

PR #789, opened 2020-05-29
Splitting metanoteinfo into metanoteinfo and intervalinfo

thenerdie:
This was something Mina suggested and I attempted to do

### DIFF
--- a/src/Etterna/Globals/MinaCalc.cpp
+++ b/src/Etterna/Globals/MinaCalc.cpp
@@ -1064,6 +1064,169 @@ is_alternating_chord_stream(const unsigned& a,
 #pragma endregion
 
 #pragma region new pattern mod structure
+// accumulates info across an interval as it's processed by row
+struct itv_info
+{
+	bool dbg = false && debug_lmao;
+	
+	int seriously_not_js = 0;
+	int definitely_not_jacks = 0;
+	int actual_jacks = 0;
+	int actual_jacks_cj = 0;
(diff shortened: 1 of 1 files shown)


nico-abram:
Could you give some explanation of what this is doing?

How much have you tested this? (to check that there are no performance regressions and that the calc outputs are the same)

Also, is this not a leak? https://github.com/etternagame/etterna/pull/789/files#diff-1ab5577893ea1c66bdc63235e181d80cR1274 where is this getting deleted?

If you're aware of what it is, have you considered using a unique_ptr there? Furthermore, could you explain why that needs to be a pointer?

thenerdie:
Yes, this is my first one.

Mina didn't want to do this so I did it instead. I have tested this using an sln build in Visual Studio 2019.

As for the leak, I'm not sure if that's actually a leak. Also, as for the pointer thing, I'm new to C++ in general so I wouldn't be able to answer that question to be honest.

About the commit message? I don't see a way that I can edit the message without using the GitHub CLI, and that's not going to work because my client is being an AUTIST right now

thenerdie:
Ok so I went in and got rid of all the pointers there was no need for them

problem solved

thenerdie:
also i keep forgetting to write commit messages .-.

thenerdie:
I'm considering closing this until Mina finishes his work. The pointers appeared back into the commit because it got overwritten by an older one.

PR #791, opened 2020-05-31
Fix RageTimer integer overflow that had been causing macOS black screen bug.

bluebandit21:
This PR fixes the incorrect std::chrono math in EventImpl_Pthreads::Wait(RageTimer* pTimeout) caused by jameskr97's commit. It also fixes the root of the macOS black screen bug, which had been that a RageTimer contained an integer overflow when attempting to convert the unix timestamp second count into microseconds.


nico-abram:
Looks good to me

@jameskr97 would you mind confirming that this keeps the issue fixed on your machine?

nico-abram:
Oh, also, @bluebandit21 , could you change the target of this pull request from master to develop? We usually use master just for releases and develop is the development branch

bluebandit21:
Sorry, done.

bluebandit21:
Uh oh. That merge borked something. Let me fix it...

bluebandit21:
Ok, so the current build system places Etterna.app inside of a "Debug" folder off of the root directory. Since the app uses relative directories to look for everything, it doesn't work until moved back to the root. So, I guess someone broke the build system. ~(-_-)~

nico-abram:
Thanks for mentioning that!

Hidden (jameskr97 on github) commited a fixed today for it: https://github.com/etternagame/etterna/commit/b81e0db256c8b33a5073a979e8693818f514b727 

Hidden's the only active dev who does anything on mac currently as far as I know so it's probably not as well maintained as windows support. He rewrote our cmake and knows the most about our build system

Since you've been interacting a good amount on github, I feel I should mention a lot of our conversations happen in the development discord (An invite is linked in the readme, the discord badge that says Etterna Dev Group, anyone can join), in case you want to join.

jameskr97:
I was able to succesfully compile this on macOS High Sierra and Mojave, and neither version experience the bug at hand. I'm going to approve this PR. @bluebandit21 Did more research into getting this working correctly compared to my quick fix.

Merge when ready.

PR #795, opened 2020-06-03
Refactor MusicWheel::FilterBySkillsets to fix several bugs and make it easier to mantain in the future

### LINKED ISSUE #781: Filters within the dance game mode incorrectly include songs with solo mode charts matching the filter
**Describe the bug**
When filtering songs in the dance game mode, songs which have solo maps which match the search parameters will be included in the results, even if none of the 4key maps of the song  meet them.

**To Reproduce**
1) Find a song which has charts in both solo and dance
"smooooch" from the Dark Chancellors pack is an example of one such song.
2) Select dance as the game mode
3) Apply a filter which applies to a solo chart of the song, but does not match any of the 4key maps for it. 
For the above example, applying the filter of 19-20 stamina will result in the song showing up, despite having no dance charts matching the query.

**Expected behavior**
Filters should only return songs containing charts from the current game mode meeting its parameters.
**Desktop (please complete the following information):**
 - OS: MacOS Mojave 10.14.6
 - Version: commit dbac6cf47a864dc54d7812eb018b87d4e20a3154 (built from source)
 - 64bit

**Additional context**
Oddly enough, the reverse isn't the case, as filters in the solo game mode only match solo mode charts. 
I would have to assume this is remnant behavior from before solo and 4key were split.

### LINKED ISSUE #794: Filters only look at the highest MSD chart of a song for each skill type
**Describe the bug**
Filters will not include songs that have matching charts if the matching chart within the song does not contain the highest-rated MSD of the song's charts for the filter's skill type.

**To Reproduce**
1) Find a song with multiple charts. POTATOE from Bursts of a Teitoku is one such song.
2) Apply a filter which matches a chart from the song that is *not* the highest difficulty of the song's charts.
In the above example, applying a filter of "Overall 22-23" should include the song, because it's hard chart matches this query, but does not.

**Expected behavior**
Filters should include songs in the results if any of the song's charts match the filter parameters, not only just the highest-MSD one.

**Desktop (please complete the following information):**
 - OS: MacOS Mojave
 - Version: Commit c258f715876e37759f3987bccbd9ba65c11d12bd (develop branch)
 - 64bit

**Additional context**
This is due to the way filtering is implemented in 
`MusicWheel::FilterBySkillsets(vector<Song*>& inv)`
I'm currently trying to fix this issue.

bluebandit21:
This refactor accomplishes many things:
1) Filter logic has been cleaned up and duplicate code has been condensed.
This should make it significantly easier to modify filter logic in the future to fix issues.
2) Highest Skillset should now work as intended with filters.
Earlier, there were two issues. First, in the old code, the continue statement after  
after checking if a given skill was the highest in a chart would continue the ratemod loop, not the skillset loop.
Secondly, in `Song::IsSkillsetHighestOfAnySteps`,
 "Overall" was included as a result, which would always (naturally) be the highest. 
By setting a boolean flag inside the function to exclude "Overall" as an option in a sub-function, 'Highest Overall' now works as intended.
(Prior, 'Highest Overall' only worked if "Overall" was the only filter active)
3) Filter logic has been moved into a new function Song::MatchesFilter, which allows filters to easily look through all charts of a song and also only sort through certain chart types (currently unimplemented). This fixes #794.
4) Filters only consider charts that have the styles of the current game mode. 
(Using `GAMEMAN->GetStepsTypesForGame`). This fixes #781.

Close: #794
Close: #781

Note: People other than me should check to make sure that there aren't any (new) filter bugs 
from this PR. I'm fairly certain that there aren't, and I've done extensive testing without issues, but I have rather significantly modified the internals of the filter.

Other Note: It made sense to me to put the filter logic into a Song function. However, this could instead easily be a subfunction of MusicWheel. I think the logic makes more sense to be under Song, since it's directly relating to the logic of a song and not the MusicWheel itself, but this could be moved if desired.

### DIFF
--- a/src/Etterna/Actor/Menus/MusicWheel.cpp
+++ b/src/Etterna/Actor/Menus/MusicWheel.cpp
@@ -732,10 +732,8 @@ MusicWheel::FilterBySkillsets(vector<Song*>& inv)
 			float ub = FILTERMAN->SSFilterUpperBounds[ss];
 			//If either bound is active, continue to evaluation
 			if (lb > 0.f || ub > 0.f) {
-				float currate = FILTERMAN->MaxFilterRate + 0.1f;
-				float minrate = FILTERMAN->m_pPlayerState->wtFFF;
-				do {
-					currate = currate - 0.1f;
+				for(float currate = FILTERMAN->MaxFilterRate;
+					currate < FILTERMAN->m_pPlayerState->wtFFF; currate -= 0.1f) {
 					if (!FILTERMAN->ExclusiveFilter) { //Non-Exclusive filter
 						if (FILTERMAN->HighestSkillsetsOnly)
 							if (!song->IsSkillsetHighestOfAnySteps(
@@ -771,8 +769,7 @@ MusicWheel::FilterBySkillsets(vector<Song*>& inv)
 							addsong=true;
 						}
 					}
-					
-				} while (currate > minrate);
+				}
 			}
 		}
 		// only add the song if it's cleared the gauntlet
--- a/src/Etterna/Actor/Menus/MusicWheel.cpp
+++ b/src/Etterna/Actor/Menus/MusicWheel.cpp
@@ -728,8 +728,7 @@ MusicWheel::FilterBySkillsets(vector<Song*>& inv)
 								 * The .01f delta is because floating points
 								 * don't like exact equivalency*/
 			if (song->MatchesFilter(currate)) {
-				// TODO: Check only relevant chart type. Currently we search
-				// all.
+				// TODO: Check only relevant chart type. (Dance, solo, etc...)
 				addsong = true;
 				break; // We don't need to keep checking rates
 			}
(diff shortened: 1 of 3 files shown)


poco0317:
refactoring musicwheel is a fearful endeavor but this can be the beginning

poco0317:
Appears to work for the most part at the moment, concerning Highest-Only filtering.
Why I say "for the most part" is that since I'm on a constantly changing version of the calc, the files showing up sometimes are not always correct. For example, I can filter by Steam 12-13 and get files which do fall into that category, but their JackSpeed may be 18... Filtering by a larger number instead reveals that they may actually be Solo files, which falls into #781 

So I would say the changes are working as intended. It would be interesting if we could also filter out specific difficulties of a chart from displaying in specific instances so that people are less confused when they filter a range and a song with multiple difficulties shows up that has only one that fits into the filter. That may get a little more involved.

bluebandit21:
Sorry about not addressing your question at the end earlier. 
I think if you're talking about selectively showing only specific charts, you have to change a whole lot about MusicWheel as a whole. (I could be wrong about that, though.)
That sounds interesting, but I think it's far beyond the scope of my meager PR.

poco0317:
It seems the logic for placing the steps into the list of difficulties available is here:
https://github.com/etternagame/etterna/blob/dbac6cf47a864dc54d7812eb018b87d4e20a3154/src/Etterna/Models/Songs/SongUtil.cpp#L1063-L1074

From like 3 minutes of looking into it. So it would probably take a minor modification here to account for the filter. But this PR as it stands is good.

PR #822, opened 2020-07-05
Add notefield columns spacing to CustomizeGameplay

caiohsr14:


poco0317:
"finished" with some effort but its not perfect
movement is not fully correct, but resizing and then resetting works fine
which is an indication that rewriting the board and the cover at the same time should fix the movement issue

Issue #832, opened 2020-07-09
Linux crash when using muFFT


poco0317:
crash goes away when commenting muFFT usage or turning off the til death visualizer.
consistent crash.

stack: 
![image](https://user-images.githubusercontent.com/2531164/87100891-1cb4ce80-c213-11ea-99eb-992d1e389c48.png)

poco0317:
#837

PR #837, opened 2020-07-12
fix mufft crash caused by unaligned input buffer

hexptr:
Fixes https://github.com/etternagame/etterna/issues/832

### DIFF
--- a/src/RageUtil/Sound/RageSound.cpp
+++ b/src/RageUtil/Sound/RageSound.cpp
@@ -355,11 +355,13 @@ RageSound::GetDataToPlay(float* pBuffer,
 			copy(pBuffer, until, back_inserter(recentPCMSamples));
 			if (recentPCMSamples.size() >= recentPCMSamplesBufferSize) {
 				auto out = static_cast<cfloat*>(fftBuffer);
-				auto n = recentPCMSamplesBufferSize;
 				auto plan = mufft_create_plan_1d_r2c(recentPCMSamplesBufferSize,
 													 MUFFT_FLAG_CPU_ANY);
-				mufft_execute_plan_1d(plan, out, recentPCMSamples.data());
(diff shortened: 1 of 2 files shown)


poco0317:
Works fine on Windows

nico-abram:
Thanks for your first pull request on this repository @hexptr !


This seems to be allocating a new buffer every time we execute the callback
Could it be rewritten to instead allocate the buffer here (And free it if it existed, same as the output buffer) and store it in the class? https://github.com/etternagame/etterna/blob/68cf405883df88a81c29b5a9727115e028f54a2a/src/RageUtil/Sound/RageSound.cpp#L780-L781

As an aside, since you're touching this. I think we can just get rid of this line https://github.com/etternagame/etterna/pull/837/files#diff-2a93c97fb469c142670452b11fddef84R365 (We're clearing the vector this writes into right after)

Neither of these is a huge deal and should not stop this from getting merged

hexptr:
Thanks for the suggestion.

I took the liberty to rewrite both input and output buffers as vectors that use a custom allocator which calls into the mufft_alloc and mufft_free functions.

Both buffers along with the FFT plan, are stored in the class now.

I added more includes and declarations into the RageSound.h header to make things work. I don't know how much we're worried about the compile speed regressing because of things like this.

nico-abram:
This looks good! I like how we also reuse the plan now. Thank you!

The need for the include in the header is unfortunate, but I don't think it's worth it to try to add some additional indirection (I'm thinking something pimpl like), the mufft header does not look particularly big (About 300 lines and mostly comments)

I compared the CI times (Except windows, which isn't finished yet) and I didn't see any noticeable changes (I imagine CI times are unreliable but it's something)

MinaciousGrace:
build times look fine to me

Issue #577, opened 2019-06-16
No replays are generated


kangalioo:
No replay data is generated from 0.66 (I think) onwards. Old replays from before 0.66 are still there and can be played, but no new replays are generated. Upon completing a chart in Etterna, a warning appears in stdout:

> WARNING: Failed to create replay file at Save/ReplaysV2/Sa64450f1d975e08845da2397876a5b832e8bb8b1"

This happens in the hotfix_66_1 branch as well.

Logs after opening Etterna, playing a single file and exiting: [Terminal output](https://pastebin.com/qX6790fc), [info.txt](https://pastebin.com/Mnf1m6m6), [log.txt](https://pastebin.com/XsSUjp4p)

**To Reproduce**
1. Play a chart
2. Go to Save/ReplaysV2 directory
3. See that no replay file was generated

**Expected behavior**
Replay files should be successfully generated and saved in the Save/ReplaysV2 directory.

**Desktop:**
 - Linux 64bit KDE
 - Etterna latest master branch (cce9555), though the issue probably started with the official 0.66

poco0317:
This may be a linux specific issue, as I cannot reproduce the problem. 
This message only shows up when std::ofstream fails to create or open the expected end result replay data. Verify that you have permission in the output directory.

kangalioo:
Write permission for the Etterna root directory, the `Save` directory and `Save/ReplaysV2` is given to the owner user, which is me. As long as Etterna doesn't switch to a different user on its own, permission shouldn't be a problem.

poco0317:
If you want to test something, try removing these lines in HighScore.cpp, or remove the false return. I would expect it to crash or do something weird if it managed to pass that point.

https://github.com/etternagame/etterna/blob/cce9555823b0e2e5fdcaa92b632f42d27a41db94/src/Etterna/Models/Misc/HighScore.cpp#L448-L451

kangalioo:
Etterna doesn't crash, nor does anything weird happen with those lines commented out. The warning message disappears (obviously) but the replay is still not written.

"log.txt" contains the following lines

> 00:40.349: saving stats and high scores
> 00:40.355: Created replay file at Save/ReplaysV2/S1d0b06bd5fb46dda56a6ef916c28e6d6e9321ead
> 00:40.355: done saving stats and high scores
> 00:40.355: Adding stats to player profile...

It seems like Etterna thinks it has written the replay file although it didn't

kangalioo:
[...]

PR #849, opened 2020-07-15
Fix a crash dereferencing m_ReplaySnapshotMap.rbegin()

hexptr:
I don't know why this crash happens, but I can consistently trigger it by pressing "up" on the evaluation screen. This patch at least fixes the crash, if not the underlying root cause.

(I suspect there's something more deeply wrong with my game -- it refuses to save any replay files, which I suspect is related.)


poco0317:
If you play on linux, replays can fail to save due to permissions or environment issues: #577

PR #859, opened 2020-07-22
Refactor SONGMAN from a singleton into a namespace

bluebandit21:
This should have minor compile-time improvements.
It should also make for clearer overall code.

Ideally, this should be the first of many such PRs to convert all of our singletons.

Note: Lua is currently broken because of this; Poco knows how to fix it.
Other Note: The physical location of code within the source file could be tidied a bit, additionally some functions are only used internally and should be moved inside of the anonymous namespace.

### DIFF
--- a/Themes/_fallback/Scripts/00 init.lua
+++ b/Themes/_fallback/Scripts/00 init.lua
@@ -25,6 +25,13 @@ Warn = lua.Warn
 -- @function print
 print = Trace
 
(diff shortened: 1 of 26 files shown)


bluebandit21:
This is now quite out of date.
Because it's old and I don't really like what I did already for this in retrospect, I'm going to close this for now.
In the future, I might try to return to this thought in a less horrifying way.

PR #860, opened 2020-07-25
Japanese Font Improvements

poco0317:
Improvements? maybe.
All I know for sure is that this fixes issues that arise when installing and using fonts that don't support these characters. They default to the fallback ones which are out of place and gigantic very often.
Font: microsoft jhenghei ui


poco0317:
In its current state, this is a start, but needs much more testing and possibly a bit of a redo with the new texture font generator.

PR #862, opened 2020-08-03
Implement ability to exclude songs based on tags

4444N:
Also adds all the GUI associated with the change to standard theme, maintains previous lua function 'FilterByStepKeys' functionality

### DIFF
--- a/src/Etterna/Actor/Menus/MusicWheel.cpp
+++ b/src/Etterna/Actor/Menus/MusicWheel.cpp
@@ -1791,15 +1791,16 @@ class LunaMusicWheel : public Luna<MusicWheel>
 		std::vector<string> newHashList;
 		LuaHelpers::ReadArrayFromTable(newHashList, L);
 		lua_pop(L, 1);
-		p->SetHashList(newHashList);
 		luaL_checktype(L, 2, LUA_TTABLE);
 		lua_pushvalue(L, 2);
 		std::vector<string> newOutHashList;
 		LuaHelpers::ReadArrayFromTable(newOutHashList, L);
 		lua_pop(L, 1);
-		p->SetOutHashList(newOutHashList);
 
+		p->SetHashList(newHashList);
+		p->SetOutHashList(newOutHashList);
 		p->ReloadSongList(false, "");
+
 		return 1;
 	}
 
--- a/src/Etterna/Actor/Menus/MusicWheel.cpp
+++ b/src/Etterna/Actor/Menus/MusicWheel.cpp
@@ -683,23 +683,37 @@ MusicWheel::SetHashList(const vector<string>& newHashList)
 {
 	hashList = newHashList;
 }
+void
+MusicWheel::SetOutHashList(const vector<string>& newOutHashList)
+{
+	outHashList = newOutHashList;
+}
 
 void
-MusicWheel::FilterByStepKeys(vector<Song*>& inv)
+MusicWheel::FilterByAndAgainstStepKeys(vector<Song*>& inv)
 {
 	vector<Song*> tmp;
-	const std::function<bool(Song*)> check = [this](Song* x) {
-		for (auto& ck : hashList) {
-			if (x->HasChartByHash(ck)) {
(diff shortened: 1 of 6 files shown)


src/Etterna/Actor/Menus/MusicWheel.cpp
```diff
@@ -1763,6 +1777,28 @@ class LunaMusicWheel : public Luna<MusicWheel>
...
+		luaL_checktype(L, 2, LUA_TTABLE);
+		lua_pushvalue(L, 2);
+		std::vector<string> newOutHashList;
+		LuaHelpers::ReadArrayFromTable(newOutHashList, L);
+		lua_pop(L, 1);
+		p->SetOutHashList(newOutHashList);
+
 		p->ReloadSongList(false, "");
 		return 1;
 	}
```
4444N, line 1804: I didn't quite understand the lua documentation on how lua call stack works, so please make sure to review these lines.

poco0317: This looks fine to me, untested, but I feel that unexpected behavior may arise when sending a non-table second parameter to this function. SetHashList may execute and then the function would error out on the next line.

4444N: This is only ever meant to be called with 2 tables as parameters, never more, never less. That's why I kept FilterByStepKeys (mostly) intact in case anyone wanted that functionality without plugging in an additional value.

poco0317: I'm inclined to guess that the parameter checks are runtime especially in those helper functions. So if you pass a nil for whatever reason, like as an accident, then the function would not exit until it reaches the `luaL_checktype(L, 2, LUA_TTABLE)` where it exits and reports the error to the game.
So what I'm saying is the parameters should be taken in and placed in those variables as the first thing rather than immediately be used and thrown away.

poco0317: However upon testing this with breakpoints real quick it seems to know that ahead of time so I'll call it good.
Didn't want to make the assumption though.

4444N: so something like
```
static auto FilterByAndAgainstStepKeys(T* p, lua_State* L) -> int
{
	luaL_checktype(L, 1, LUA_TTABLE);
	luaL_checktype(L, 2, LUA_TTABLE);
	lua_pushvalue(L, 1);
	lua_pushvalue(L, 2);
	std::vector<string> newHashList;
	LuaHelpers::ReadArrayFromTable(newHashList, L);
	lua_pop(L, 1);
	std::vector<string> newOutHashList;
	LuaHelpers::ReadArrayFromTable(newOutHashList, L);
	lua_pop(L, 1);

	p->SetHashList(newHashList);
	p->SetOutHashList(newOutHashList);

	p->ReloadSongList(false, "");
	return 1;
}
```
would be better in your opinion then?

poco0317:
```cpp
	static auto FilterByAndAgainstStepKeys(T* p, lua_State* L) -> int
	{
		luaL_checktype(L, 1, LUA_TTABLE);
		lua_pushvalue(L, 1);
		std::vector<string> newHashList;
		LuaHelpers::ReadArrayFromTable(newHashList, L);
		lua_pop(L, 1);
		luaL_checktype(L, 2, LUA_TTABLE);
		lua_pushvalue(L, 2);
		std::vector<string> newOutHashList;
		LuaHelpers::ReadArrayFromTable(newOutHashList, L);
		lua_pop(L, 1);

		p->SetHashList(newHashList);
		p->SetOutHashList(newOutHashList);
		p->ReloadSongList(false, "");

		return 1;
	}
```

4444N: OK I see so that the function doesn't call anything external before the parameters are set up.
Sure I'll change that then.

PR #863, opened 2020-08-04
Fix several compile-time warning messages to make build output more useful

bluebandit21:
The following changes were made:
Parenthesis were placed around some logical operators to remove warnings about implicit ordering.
Pragmas were placed to explicitly suppress warnings about non-literal format strings within `ssprintf`.
An incorrect format specifier was fixed within the `WERROR` macro.
An enum value that had been "added" to a string was changed to actually be appended.

This drastically reduces the number of warnings on my machine (100+ into ~15), and should generally make building information significantly more useful, as well as allowing important warnings to be seen.

### DIFF
--- a/src/Etterna/Models/Misc/GameCommand.cpp
+++ b/src/Etterna/Models/Misc/GameCommand.cpp
@@ -431,8 +431,8 @@ GameCommand::ApplySelf(const vector<PlayerNumber>& vpns) const
 			case StyleType_OnePlayerTwoSides:
 				break;
 			default:
-				LuaHelpers::ReportScriptError("Invalid StyleType: " +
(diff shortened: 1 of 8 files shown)


poco0317:
It's a step in the right direction, but in Windows GHA x99999: `  D:\a\etterna\etterna\main\src\RageUtil/Utils/RageUtil.h(550,9): warning C4068: unknown pragma (compiling source file D:\a\etterna\etterna\main\src\RageUtil\Graphics\RageTextureRenderTarget.cpp) [D:\a\etterna\etterna\main\build\Etterna.vcxproj]`

bluebandit21:
That should be fixed now.

PR #869, opened 2020-08-13
Match cursor updaterate to display refresh rate

Jolomann:
Minor tweak. Now possible thanks to https://github.com/etternagame/etterna/commit/4c5dec0687b368019b678bb3922fbeff23c2e050


poco0317:
Is there a reason we dont do `self:SetUpdateFunctionInterval( 1 / DISPLAY:DISPLAY:GetDisplayRefreshRate() )` instead?

Jolomann:
There isn't. I can change it to that if it's better

poco0317:
The have the same end effect, but using SetUpdateFunctionInterval would force the update speed to only apply to those things relevant to the update function instead of all children.
Think of what holding Tab does to the game. Holding Tab multiplies all update rates by 4. Your code multiplies the update rate of the ActorFrame by the refresh rate.

Jolomann:
Ah, might as well only update the function then

PR #882, opened 2020-08-22
💥 Improve Crash Handling

jameskr97:
# Crash Handling with Google Crashpad
The old crash handler and checkpoint systems have been completely removed, and replaced with [Google Crashpad]. Crashpad is a library for capturing and transmitting crash reports. This is a significant improvement over the previous crash handling system, as now we are able to generate a complete stack trace on every thread at the time of the crash. Crashpad also renders checkpoints unnecessary as that was an attempt to get the same information.

## Important Notes
- This PR is **NOT** to be merged until #911 has been merged. This PR was rebased to include the changes of #911, and merging this first will result in two PR's being merged at once.
- This PR is **NOT** to be merged until #917 is also ready to be merged. They are meant to be merged together, with this PR focusing on the compilation of crashpad, and the #917 improving on the integration.
- Linux builds will not come with crash handling. Since we do not distribute linux binaries, there is no way to store symbols from the generates binaries. Linux users will now be required to generate their own symbols and include them in crash reports. This process will be included in the latest documentation setup.

## PR Changes
- [x] Remove all files within `src/archhooks` and `src/arch` related to crash handling and backtracing.
- [x] Crashpad source code is added into `etterna/crashpad/crashpad`. 
- [x] Debug Symbols are generated on Github Actions and uploaded as artifacts. These are now a required part of being able to debug client-generated minidump files.
- [x] Symbols generated from the master branch uploaded to Amazon S3 for long term storage.
- [x] Allow crashpad to be disabled on developer systems. While devs may find it useful, it is primarily the job of CI to ensure that symbols are generated from the same executable that gets distributed. It is not necessary for developers to perform this regularly, and will usually have their IDE's to catch exceptions.
- [x] Checkpoints completely removed.

~~Improve documentation to better explain how Crashpad is included and compiled within Etterna, and how to compile it locally. This includes documentation on how to decode a minidump to produce the stack trace.~~ More significant documentation updates will be completed in #884.

[Google Crashpad]: https://chromium.googlesource.com/crashpad/crashpad/


poco0317:
While we are ahead I would like to warn you that replacing all `CHECKPOINT_M` with a log trace will probably inflict some type of performance hit or induce log file bloat and we need to evaluate each replacement individually.

jameskr97:
> While we are ahead I would like to warn you that replacing all `CHECKPOINT_M` with a log trace will probably inflict some type of performance hit or induce log file bloat and we need to evaluate each replacement individually.

Would changing the default log level to info work for decreasing the number of log messages?

poco0317:
It would probably reduce it to almost none

poco0317 (commented):
While this may disable crashpad, it does not allow normal building. `WITH_CRASHPAD` probably needs to be set false if python isn't found

Or we can make sure that's in the documentation and if anyone fails to build then they should be able to figure out why.

src/Etterna/Models/Misc/PlayerAI.cpp
```diff
@@ -1047,8 +1047,8 @@ PlayerAI::SetPlayerStageStatsForReplay(PlayerStageStats* pss, float ts)
 	pss->m_fLifeRecord.clear();
 	pss->m_fLifeRecord = GenerateLifeRecordForReplay(ts);
 	pss->m_ComboList.clear();
-	pss->m_ComboList = GenerateComboListForReplay(ts);
```
poco0317, line 1050: lost behavior here?

poco0317 (changes requested):

src/archutils/Win32/ddk/hidpi.h
```diff
@@ -306,8 +306,8 @@ Routine Description:
    Capabilities     a HIDP_CAPS structure
 
 Return Value:
-�  HIDP_STATUS_SUCCESS
-�  HIDP_STATUS_INVALID_PREPARSED_DATA
+�  HIDP_STATUS_SUCCESS
```
poco0317, line 309: possible encoding issues in this file?

src/Etterna/Models/Misc/PlayerAI.cpp
jameskr97: Fixed in next commit.

src/archutils/Win32/ddk/hidpi.h
jameskr97: This file was originally in the codebase with ISO-8859-1 encoding. The mojibake was caused by a bad conversion to UTF-8. I downloaded the previous file before I touched it, converted it to UTF-8, and the fixed version will be in the next commit.

jameskr97:
The above force push is a rebase.

poco0317 (approved):
It's been tested a fair bit. It does provide crash info for typical crashes. I do not get a crash dump for a heap corruption panic, interestingly. The only thing I don't like is that there is no message indicating a crash happened - the game just closes. Otherwise the dumps are informative enough to be useful to me. Regular users need not care about their contents if we can reduce the frequency of crashes in general anyways.

poco0317:
Status on this seems to be that /Logs/ does not generate from CI build but do from dev environment. Crashdumps do generate on x64 and x86.

poco0317:
As of now, this has reached a satisfactory state and can be merged when the next-in-line #917 is ready (it provides minor additional fixes to code in this PR)

Issue #881, opened 2020-08-22
Maybe change MinaCalc Chisel implementation to use binary search, for performance


kangalioo:
The [`Chisel`](https://github.com/etternagame/etterna/blob/0b7a28d2371798a8138e78e5789d0014b16b4534/src/Etterna/MinaCalc/MinaCalc.cpp#L523) function in MinaCalc, which takes up a big part of execution time as far as I know, is implemented in a potentially inefficient way. Performance would probably increase if the search algorithm was implemented using binary search instead.

I didn't thoroughly test the actual performance impact yet; although I did test how much it affected performance to introduce binary search in a Rust port of one of the Aggregate* functions - they share the same algorithm as Chisel in MinaCalc. The performance gain was +30%.

The downside of binary search over the current implementation is that binary search can only search a certain range, so `Chisel` wouldn't be able to return any value beyond, say, 163.84, or 327.68 (depending on the number of iterations and resolution). However, I think this is not that big of an issue because the calc is only designed for a certain scope of values anyways.

poco0317:
#883 reopen or discuss if further changes needed

PR #883, opened 2020-08-23
Refactor MinaCalc's Chisel function to improve performance

bluebandit21:
I profiled both the new and old calc to compare performance by deleting cache.db and timing from game-start to the first time I saw the menu screen using XCode's Instruments tool.

|Measurement|Old Calc | New Calc|
|---|---|---|
|Relative Chisel weight to total execution time|12%|10%|
|Total CPU-seconds spent in Chisel<br>(Slighty unreliable metric)|58.36|46.45|
|Total run time (CPU-Mins)<br>(Highly unreliable metric)|8.08|7.67|

Although this improvement is small, the new code is cleaner and easier to understand than the old code was, making this a change with no downsides.
There was zero difference in calculated values between the old calc and the new one with my changes, using the Calc test xml.
Related to #881 -- the first iteration is unimproved, but the 2nd-7th should take 50% less time on average.

### DIFF
--- a/src/Etterna/MinaCalc/MinaCalc.cpp
+++ b/src/Etterna/MinaCalc/MinaCalc.cpp
@@ -943,7 +943,7 @@ MinaSDCalcDebug(
 	}
 }
 
-int mina_calc_version = 438;
+int mina_calc_version = 439;
 auto
 GetCalcVersion() -> int
 {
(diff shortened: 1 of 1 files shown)


poco0317:
My scuffed profiling (run CalcEtternaMetadata 1000 times on the same file and measure the time taken)
Measured 3 times and took the average of the results.

|Song|438 Calc (ms) | 439 Calc (ms)|
|---|---|---|
|Game Time|19896.00|17424.33|
|Conspiracy of Silence|128951.00|116216.67|

Appears to be 10-15% faster.

bluebandit21:
Hey, that's a win in my book ;D

PR #896, opened 2020-09-09
Update README.md to include MacOS installation procedure

bluebandit21:
This is unlikely as-is to change between releases, so it doesn't make sense
to copy it verbatim every single release while having inaccurate info visible.


bluebandit21:
Force push because I mis-capitalized macOS as MacOS.

PR #897, opened 2020-09-11
Tidy up building docs a bit, add arch dependencies

Ulti-FD:


jameskr97:
Why are arch dependencies being added? I thought only Fedora and Debian/Ubuntu were being supported.

Ulti-FD:
Works just fine for me, have built on Manjaro and stock Arch. Why not make the info more accessible if it works anyways?

Ulti-FD:
Packaging is a non-issue for Arch-based stuff at this point, Mado is currently maintaining a PKGBUILD for the AUR. I imagine no effort will be needed on your part. Unfortunate debian-based distros don't have something similar though.

PR #905, opened 2020-10-04
Add virtual default destructor to Core::ILogger so child destructors …

bluebandit21:
A virtual destructor was added, as when a base class does not have a virtual destructor, calling a derived destructor in the form of...
```
Base *base = new Derived();
delete base;
``` 
... will cause undefined behavior. This PR will guarantee default destruction.

### DIFF
--- a/src/Core/Services/ILogger.hpp
+++ b/src/Core/Services/ILogger.hpp
@@ -18,6 +18,9 @@ namespace Core {
 class ILogger {
 
 public:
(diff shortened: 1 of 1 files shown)


jameskr97:
Hope you don't mind me changing the description, I just wanted it to be more clear as to why this PR is a good idea. Will merge after CI.

bluebandit21:
Of course not! 
Thanks for making it clearer

Issue #908, opened 2020-10-19
Invalid Step Files Should Not Be Able to Crash Etterna


bluebandit21:
A number of issues have arisen due to invalid step files crashing Etterna. 
One particularly notable example is that any files with zero-length holds (see: Osu maps) will crash Etterna and leave the person to have to hunt down which file has them.

Ideally, we should not crash Etterna under *any* invalid input, and should instead just "skip" that file, printing an unhappy warning message in the logs.

At a minimum, if loading a file crashes Etterna in any way, anywhere, the last output to the log should be the name of the file, as well as some basic debug info if possible to aid people trying to find the issue.

In general, Etterna should be hardened against invalid user input of any kind, failing gracefully whenever possible and crashing with useful debugging info otherwise.

I'm reasonably certain that an issue a user is experiencing right now is caused by a malformed dwi file, but the crash output is a random access violation, which is near-useless for troubleshooting. This is why I made this PR in the first place.

poco0317:
One point of exception handling for song load in general, assuming exceptions just propagate through the call stack without any help, would be handling it in `Song::LoadFromDir` (loader LoadFromDir) and `Steps::GetNoteDataFromSimfile` (loader LoadNoteDataFromSimfile)
That probably would catch most io/load related issues.

PR #909, opened 2020-10-19
Replace stof with stod (truncating result to float) to prevent low-value crash

bluebandit21:
Extremely small BPM values being passed into the OSU loader could cause
 Etterna to crash by making stof go out of range by passing a value in too
 small to be representable by a float.

This change sidesteps this issue by instead using stod, and allowing the
 result to be truncated to a float (truncating unrepresantable values to 0).

This can still technically crash, but you'd need to pass in even more
 ludicrously small values to do so.
In the future, it might make sense to wrap this in a try-catch block and
 catch the std::out-of-range exception to continue cleanly.

Note: I'm leaving this draft for now to let someone more knowledgeable than me make sure it's ok.
Notably, the second value can be exactly 0; I don't know if that's potentially harmful. 
I tried playing a file that previously crashed for the above reason, and it seemed to play correctly.

### DIFF
--- a/src/Etterna/Models/NoteLoaders/NotesLoaderOSU.cpp
+++ b/src/Etterna/Models/NoteLoaders/NotesLoaderOSU.cpp
@@ -149,7 +149,7 @@ OsuLoader::SetTimingData(map<string, map<string, string>> parsedData, Song& out)
 		auto values = split(line, ",");
 
 		tp.emplace_back(
-		  std::pair<int, float>(stoi(values[0]), stof(values[1])));
+		  std::pair<int, float>(stoi(values[0]), stod(values[1])));
 	}
 	sort(tp.begin(),
 		 tp.end(),
@@ -353,8 +353,9 @@ OsuLoader::LoadNoteDataFromSimfile(const std::string& path, Steps& out)
 {
 	RageFile f;
 	if (!f.Open(path)) {
-//		LOG->UserLog(
-//		  "Song file", path, "couldn't be opened: %s", f.GetError().c_str());
+		//		LOG->UserLog(
+		//		  "Song file", path, "couldn't be opened: %s",
+		// f.GetError().c_str());
 		return false;
 	}
 


poco0317:
I think a solution which is less likely to cause exception related issues is replacing it with `strtof()`
This requires changing the code to `... strtof(values[1].c_str(), nullptr)`
The function ignores the second argument if it is null. Usually it is used to provide a position to continue parsing chars in a sequence. The reason we would use this function is that it doesn't throw an out of range exception. Instead it returns, in this context, `HUGE_VALF` or `0`. In testing on the same file, it ended up returning 0, possibly due to failed parsing rather than out of range (but that much is a guess from me). It's also a positive to keep the type consistent.
If this causes more issues down the line we can handle them within the TimingData code to make it more resilient probably.

bluebandit21:
Oh, that's a much better fix than mine was.
I'll make the change later today and update the PR.

On a related note, might it be worth looking across the codebase for other instances of stof that we might want to do the same change on? (Especially for the song loaders)

poco0317:
`stof()` has uses in NotesLoaderOSU, HighScore (replay loading), and debug related uses in ScreenSelectMusic.
`stoi()` has more uses.
Not all are completely necessary to replace. Exception handling might even be better, though. I guess it's a matter of how consistent you want to be in the usage across the codebase between stof and strtof.
Be aware that stoi/stof/etc are c++ std functions while strtof/etc are c std functions which probably explains why they have fallbacks instead of throwing an exception.

bluebandit21:
Actually, alternative idea for fixing this particular instance.
What if we kept stof and implement proper exception handling right around it?

That might be a better idea if we eventually want to do what I suggest in #908 and implement some sort of exception-based framework around loading any song.

poco0317:
Yeah my initial thought was to handle the exception here and set bpm to 0. I deferred to strtof since it just worked immediately. I'm unsure of what happens if we set bpm to essentially infinity, but the game handles 0 bpm. It wasn't clear to me how to figure out which side of the "out of range" the number falls on so that we could either use 0 or infinity.

poco0317:
One point of exception handling for song load in general, assuming exceptions just propagate through the call stack without any help, would be handling it in `Song::LoadFromDir` (loader LoadFromDir) and `Steps::GetNoteDataFromSimfile` (loader LoadNoteDataFromSimfile)

PR #917, opened 2020-10-30
🛠 Improving Arch Specific Code

jameskr97:
# Better ArchHooks
ArchHook focuses on platform-specific code, and how it can be used throughout the codebase in a seamless manner as possible.

It's currently set up with a base class of functions to be implemented, with implementation decided at compile time. Now, with a stable build system, the necessary implementation can be decided before compilations start, greatly clarifying what files are included. The format was also switched from subclasses to a single namespace, with an implementation per operating system. Some functions that were considered "platform specific" such as `SetHasFocus` and `SetToggleWindowed` were moved to `GameLoop.{h, cpp}`, as while they were triggered by platform specific functions, the operations themselves, are not platform specific.

## Important Notes
- This PR is **NOT** to be merged until #882 has been merged. This PR was rebased to include the changes of #882, and merging this first will result in two PR's being merged at once.

## PR Changes
- [x] Create `Core/Arch/Arch{Mac,Win,Unix}.cpp` for each platform.
- [x] Complete ArchMac implementation
- [x] Complete ArchWin implementation
- [x] Complete ArchUnix implementation
- [x] Add `ghc::filesystem` for cross-platform path management.
- [x] Add `Core::AppInfo::CRASHPAD_HANDLER_EXE` and appropriate logic to simplify crashpad integration for development and distribution.
- [x] Organized `GameLoop.h`


poco0317:
Request: implementations to open a folder in the system file explorer. Not exactly meant for lua accessibility -- putting something like that into the debug menu would be useful for charters or people looking to explore the folders for particular songs. All you have to provide is the working ArchHooks function and I can throw it into the debug menu.

poco0317:
In current state seems good and it at least runs which means it isn't broken. Because of the default log level change at some point, many logs are gone except the startup info. Will need to look into this either before or after merge along with an audit of all logs so we can start getting really informative stuff.

PR #931, opened 2020-12-12
Update microsoft/ms-build from v1.0.0 to v1.0.2

4444N:
This fixes the currently broken Windows CI by updating https://github.com/microsoft/setup-msbuild to v1.0.2 which removes the deprecated usage of add-path in CI.

https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -79,7 +79,7 @@ jobs:
           path: dxsdk
 
       - name: Add MSBuild to PATH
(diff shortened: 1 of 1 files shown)


poco0317:
Conflicts with set-to-merge #917 (exact same change)
https://github.com/jameskr97/etterna/blob/17eb3178f8fffdc2d657c3692e8220c042a89ec0/.github/workflows/ci-workflow.yml#L129

PR #941, opened 2020-12-30
Add C++ Backend For New Filter Option -- Chart Clear %

bluebandit21:
This adds in a new filter option -- a clear % range.

Charts are only matched in the filter if the player's best score percent on a chart falls between the lower and upper bounds.
This also respects the SSRNormSort option -- if it is enabled, that clear % is used, if it is disabled, the % of the judge used when the song was played is used instead.

This also introduces a small filter change -- certain filter options (currently duration and clear %) must be matched, even in a non-exclusive filter.
This is done because no-one wants to filter for "Charts that are (between 21-23 JS), (between 20-22 HS), or (have been cleared with at least 60%)", they want "Charts that are ((between 21-23 JS) or (between 20-22 HS)) *and* (have been cleared with at least 60%)".

The easiest way to test this new filter option is to modify `FilterManager::ResetSSFilters()` by temporarily adding in lines like the following at the end for testing:
```	
SSFilterLowerBounds[NUM_Skillset+1] = 70;
SSFilterUpperBounds[NUM_Skillset+1] = 90;
```

### DIFF
--- a/src/Etterna/Actor/Menus/MusicWheel.cpp
+++ b/src/Etterna/Actor/Menus/MusicWheel.cpp
@@ -286,8 +286,9 @@ MusicWheel::SelectSongOrCourse() -> bool
 		}
 	}
 
-	Locator::getLogger()->trace("MusicWheel::MusicWheel() - No selectable songs or courses "
-			   "found in WheelData");
+	Locator::getLogger()->trace(
+	  "MusicWheel::MusicWheel() - No selectable songs or courses "
+	  "found in WheelData");
 	return false;
 }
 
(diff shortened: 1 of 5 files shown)


bluebandit21:
TODO: The C++ front end for Til Death has not yet been made, so it is not yet possible to use the new filter range.

I will implement that functionality in a separate PR.

poco0317:
I finished the UI for this since it took a while and after looking at it I figured it would be an actual nightmare to deal with
testing took place in that UI
UI is restricted to number entry only, allowing input of between 1% and 99.999%
the decimal is automatically added after entering 2 numbers
so 1-10 are single digit percentages only (who is really going to want to filter with a bound in that range anyways)

PR #951, opened 2021-02-13
Add static linked openssl notice for arch

g-robertson:
Arch Linux's package manager uses dynamically linked libraries for OpenSSL, so I added a notice for arch in building.md to note that the project forces statically linked openssl and how to disable it.


poco0317:
as a start it is worth to note that on the develop branch as it stands there is nothing on line 58 in cmakelists.txt

you dont want to reference specific lines like that
also wouldnt it be better to run `cmake -DOPENSSL_USE_STATIC_LIBS=off` rather than modifying files

PR #952, opened 2021-02-13
Change OpenSSL to dynamically link when not on windows

g-robertson:
This allows for security updates related to OpenSSL to come faster for Unix systems that have dynamically linked OpenSSL as well as allowing building on Unix systems that do not have statically linked OpenSSL.


g-robertson:
I've considered that dynamically linking OpenSSL on Mac's might imply that they need to have OpenSSL installed themselves in order to run the generated binaries

PR #958, opened 2021-03-14
Fix 404 error parsing of top/25 API answer

leovilok:
In the JSON returned by the API, "errors" is an array, not an object: this lead to a failed assert in the d["errors"].HasMember() call.

I don't know if the other "errors" that can be returned in other cases should be arrays too.

I got this crash while loging in, as a new user with no score on etternaonline, this probably affects every new user.

### DIFF
--- a/src/Etterna/Singletons/DownloadManager.cpp
+++ b/src/Etterna/Singletons/DownloadManager.cpp
@@ -1977,8 +1977,9 @@ DownloadManager::RefreshTop25(Skillset ss)
 	auto done = [ss](HTTPRequest& req, CURLMsg*) {
 		Document d;
 		if (d.Parse(req.result.c_str()).HasParseError() ||
-			(d.HasMember("errors") && d["errors"].HasMember("status") &&
-			 d["errors"]["status"].GetInt() == 404) ||
+			(d.HasMember("errors") && d["errors"].IsArray() &&
+			 d["errors"][0].HasMember("status") &&
+			 d["errors"][0]["status"].GetInt() == 404) ||
 			!d.HasMember("data") || !d["data"].IsArray()) {
 			LOG->Trace(
 			  ("Malformed top25 scores request response: " + req.result)


poco0317:
The thing with this is that the "errors" value returned can either be an Object or an Array of Objects (which look the same as the single Object)
Our private API documentation states this for a 404 on the endpoint this PR mentions, for example:
```
{
  "errors": [
    {
      "status": 404,
      "title": "User not found",
      "detail": "This username was not found"
    }
  ]
}
```

And some methods in downloadmanager handle either one of these cases. The reason why? I dont have the answer.
We are looking into rewriting a large portion of downloadmanager soon since we will likely be changing all of the API endpoints for a site rewrite as well as moving downloadmanager onto its own thread.
If you don't want to do any further investigation, this can be merged as is. The reason it was fine before this change anyways is that the client shouldn't be trying to acquire a nonexistent profile anyways since in order to log in and use this method the account has to exist. But that's not a proper expectation.

leovilok:
OK I guess I'll need to make another PR to target it to the develop branch.

I wanted to avoid having to actually clone the repo so I copied the fix I wrote on my copy of the 0.70.3 source code into the github web UI (this part of the file doesn't seem to be different in any recent version anyways).

I won't fix the rest of the file if you say that the API will change anyways. I can also keep the fix for myself if the download manager is to be rewritten before the next release anyways ... it works for me now after all :wink:.

leovilok:
> As for what you said about logging in as a new user with no scores, that shouldnt crash. I'm interested in if this is actually the reason you crashed or something else is at work there.

I traced it in gdb, so the crash definitively comes from this precise error check (the HasMember() asserts that what it's working on is an object (IsObject()), than this creates an exception that triggers the crash handler, and there is actually a crash in the crash handler itself).

poco0317:
I was wondering because plenty of people have logged in with no scores on their account before and have even recently discovered that it is possible to log in ingame before activating your account via email

leovilok:
Maybe there were an issue with my account: now that I have a few (miserable) scores in, even unpached Etterna doesn't crash anymore.

poco0317:
superceded by #959

PR #959, opened 2021-03-14
Fix 404 error parsing of top/25 API answer

leovilok:
In the JSON returned by the API, "errors" is an array, not an object: this lead to a failed assert in the d["errors"].HasMember() call.

I got this crash while loging in, as a new user with no score on etternaonline, this probably affects every new user.

The handling of API returned "errors" is inconsistent in the download manager at this point, this should probably get worked on in the future.


poco0317:
related #958

PR #966, opened 2021-03-29
Make external builds not break from local CMakeLists

g-robertson:
This commit **should** allow using Etterna's CMake structures externally for projects that depend on certain library-like functionalities within Etterna like NoteLoader or MinaCalc without removing the ability to build or functionalities of the CMakeLists altered.

The CI passes, however I have only tested this change in linux

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,7 +131,7 @@ add_subdirectory(src/Core)
 
 ## The source_group line creates the full visual studio filter layout
 get_target_property(sources Etterna SOURCES)
-source_group(TREE ${CMAKE_CURRENT_SOURCE_DIR} PREFIX "Etterna" FILES ${sources})
+source_group(TREE ${CMAKE_SOURCE_DIR} PREFIX "Etterna" FILES ${sources})
 
 # Static Analysis
 include(CMake/Helpers/StaticAnalysis.cmake)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -131,7 +131,7 @@ add_subdirectory(src/Core)
 
 ## The source_group line creates the full visual studio filter layout
 get_target_property(sources Etterna SOURCES)
-source_group(TREE ${CMAKE_CURRENT_SOURCE_DIR} PREFIX "Etterna" FILES ${sources})
```
g-robertson, line 134: It is very worth noting I have not seen how much this changed line affects Visual Studio

g-robertson: Locally CMAKE_CURRENT_SOURCE_DIR and CMAKE_SOURCE_DIR are the same value so this change should do nothing to normal builds

jameskr97:
Hey there. I'm a little unsure with what you're trying to do here. What's the use case?

nico-abram:
I think they're trying to use our cmake files to build minacalc and possibly notesloaders as they mentioned (The classes to parse .sm files, for example) as a library, for use in other projects (Like python through ffi or another native project)

There are some precedents for people doing that kind of thing, iirc rop does that on EO, and I think a couple people have built python tools to make graphs from calc output. There's also this WASM thing that graeme made: https://github.com/graemephi/seeminacalc/

poco0317:
This PR stems from a conversation on EO on 3/27 about making an EO-like site or service to search for songs using an advanced filter utilizing the calc. Similar to some of our goals for EO2

jameskr97:
Could we extract the parts of the game which are more "library like" into a separate repo?

poco0317:
MinaCalc is the central item worth doing that do, but it also uses some game source items such as RageFile. Omitting these references, it can be compiled on its own like is already done for EO

poco0317:
works in visual studio

PR #974, opened 2021-04-17
Update luajit

nico-abram:
I'm not actually sure if we want to do this or not, skimming the luajit commit history it seems there's been a few bug fixes and at least one perf improvement

Figured I'd open this PR which seems to mostly work (I've only tested it on windows for a bit, could play fine, haven't checked macos or linux builds other than getting them to build with github actions) so at least we can discuss if we want this and so the changes are available if we decide to do it


poco0317:
Needs testing

poco0317:
superceded by #1010

PR #992, opened 2021-05-19
Add macOS installer for game automatically generated by GHA CI

bluebandit21:
This should fix all of the macOS installation problems.

To use, the user just has to double-click the GHA dmg to extract it, after which there will only be an "EtternaInstaller.app" inside.
When that is run, the user is prompted to choose an install location for Etterna.

The Etterna install made by it will work without any of the gatekeeper / app translocation issues that have been plaguing us.


bluebandit21:
Closing for now because #1001 is a better solution.
If the above fix ever breaks in the future from using a private API, this can be reopened.

PR #996, opened 2021-05-31
🖥️ Improved Window System & 🧹 Minor Organization

jameskr97:
# GLFW to replace LowLevelWindow

This PR removes LowLevelWindow for GLFW for the Windowing system only. GLFW will eventually come to handle all Input and Controllers, more changes must be made between how tightly bound the game is to the framework.
This PR started out as reworking the OpenGL code. Since that was the originally intention `RageDisplay_OGL` was changed to organize the source code with labels, and glew was replaced with glad. Any commit prefixed with `Dead Code` had code that wasn't used anywhere on any platform. This PR later became focused on the windowing system, since it was a large piece that could be easily isolated.

## Changes Include
- Organizing `StepMania.{cpp,h}`
- `RageDisplay_OGL` organization
- `glew` replaced with `glad`
- `LowLevelWindow` replaced with `GLFWWindowBackend` and `IWindowBackend`


src/RageUtil/Graphics/Display/RageDisplay.cpp
```diff
@@ -155,217 +179,13 @@ RageDisplay::SetVideoMode(VideoModeParams p, bool& bNeedReloadTextures)
...
-{
-	std::string s;
-	// If FPS == 0, we don't have stats yet.
-	if (!GetFPS())
-		s = "-- FPS\n-- av FPS\n-- VPF";
-
-	s = ssprintf("%i FPS\n%i av FPS\n%i VPF", GetFPS(), GetCumFPS(), GetVPF());
-
-	//	#ifdef _WIN32
-	s += "\n" + this->GetApiDescription();
```
poco0317, line 213: Since this PR keeps both implementations of display renderer functional, losing this line has removed useful debug information

poco0317 (changes requested):
currently capped at 144fps with no functional input on windows using opengl renderer

poco0317:
In current state: d3d and opengl are functional and take input. Opengl is capped at 144fps due to vsync reasons (we had a solution to this and I think the code was left out)

Opening the game in opengl and closing it causes the vsync preference to turn itself on mysteriously (with and without the vsync code added). Opengl itself performs surprisingly bad when certain things are happening on screen. More investigation and testing required here, I haven't done much. It seems that changing the texture resolution to the lowest setting in opengl causes the game to run at an extremely low framerate, and moving the mouse around a lot in opengl causes impressive fps dips.
Also still missing the information mentioned in the requested changes above.

Again, needs thorough testing and comparison between master opengl and this pr. d3d seems good

jameskr97:
Closing for #1013. This one is doing too much. The changes will still be added in different PRs.

PR #998, opened 2021-06-13
Show number of played songs per pack in song wheel

kangalioo:
In the song wheel, the number that previously showed the number of songs in a pack has been expanded to show the number of played songs as well. The numbers are faintly colored red or green depending on percentage of played songs.

This can be useful to find packs you once downloaded but never really played. It can also be useful for people who like to play every song of every pack, as a sort of progress indicator.

Should this change be gated behind an in-game setting?

Screenshot_20210613_185850


poco0317:
the more attributes about pieces of the musicwheelitems we modify the less lua is free to modify on its own

some players do not wish to play packs to completion and this feature is not useful to them, so should be put behind a preference

kangalioo:
> the more attributes about pieces of the musicwheelitems we modify the less lua is free to modify on its own

Is there a better way to implement this which gives Lua more freedom?

> should be put behind a preference

Sure, I'll look into it and hopefully figure out how to do it

poco0317:
to give lua or the overall theme more freedom would be to allow disabling the feature

kangalioo:
While playing I also noticed this fun quirk when looking at search results 
Screenshot_20210614_232640

Not sure what the best way to resolve this is:
- `9/1`: keeping it like this? As in "you've played 9 songs in this pack of which you are seeing 1 result"
- `9/40`: showing full pack size even in search results
- `1/1`: limit the shown number of played songs to the displayed subset of pack songs

poco0317:
the number should be easy to explain, and in this case 1/1 is the only easy one to explain

bluebandit21:
You might already know this, but the fix is pretty easy -- just swap out the `GetChartsOfCurrentGameMode()` with `GetChartsMatchingFilter`

bluebandit21:
Actually, wait, I don't think that function respects search results too -- sorry for the bad comment 0_0

kangalioo:
Is there something left that needs to be improved/implemented? The feature is now off-by-default and togglable in Advanced Options. Numbers are displayed correctly even with filters or search enabled

MinaciousGrace:
is there any reason the counting is done outside the bool check? and is there any appreciable overhead on it or is it basically instant, not that i particularly care if it's option locked, but more curious

kangalioo:
> is there any reason the counting is done outside the bool check?

No, just a silly oversight.

> and is there any appreciable overhead on it or is it basically instant, not that i particularly care if it's option locked, but more curious

I personally don't notice a difference when comparing scrolling through vanilla 0.70.3 and my num-played branch.

poco0317:
it runs once for every section the wheel has loaded every time you move the wheel which can kind of get really bad

if code isnt going to get used due to a preference it should be hidden behind the preference

kangalioo:
Yes I already fixed the code, just didn't push yet.

Also, there is a small display quirk in alternative sort modes. Need to figure that out before merging

kangalioo:
This branch now compiles successfully and the feature seems to work, even with filters+search+alternate sort enabled.

PR #1006, opened 2021-07-07
the ThemeStuff™

Ulti-FD:
A collection of theme and other misc changes intended to be merged for 0.71
Mostly lua edits, c++ edits are few. Til Death changes for the most part
Goals:
- Visual clarity
- Distinction of non-button and button texts
- Easier to understand UI here and there
- Functional improvements, more customization options
- Ports of some of the features/enhancements from scwh/rebirth
- Error fixes

Hopefully these commits are done in a way that makes putting them in the changelog not a nightmare
I have a few more things I want to get in this before I undraft, shouldnt take me more than two days
Thanks for the help as always poco


Ulti-FD:
I have other things I want to do but I really should call it, I'm running low on steam
There's always next update, this is plenty
I'm not expecting you to read through every line but maybe do skim a bit and check I didn't do something aweful

PR #1018, opened 2021-07-31
Platform Unix: escape the URL passed to xdg-open

staticssleever668:
This fixes some shell errors, for example when there is an apostrophe (`sh: 1: Syntax error: Unterminated quoted string.`) or a parenthesis (`sh: 1: Syntax error: "(" unexpected`) in the URL.


jameskr97:
👍

PR #1022, opened 2021-08-02
Rename libMAD file VERSION to avoid issues with C++20-compatible comp…

bluebandit21:
…ilers

C++20 includes the <version> header, which will incorrectly pull in files named VERSION within header include paths.

As a result, compilation of Etterna will fail on C++-20 compatible compilers because <stddef> will include random garbage instead of the proper <version> header.

Renaming libMAD's VERSION to VERSION.txt prevents this behaviour.

However, this PR does not fix the same issue caused by the VERSION file generated by FFMPEG.

See: https://bugs.llvm.org/show_bug.cgi?id=42540


bluebandit21:
This is why programming is a fun time.

jameskr97:
Does this happen from current develop, or when Etterna is switched to compile in C++20?

bluebandit21:
Current develop building on latest XCode on latest Big Sur without modifying any settings.

bluebandit21:
Actually, that last bit isn't *quite* true -- 
As mentioned on the Dev Discord, I had to pass in the `-T buildsystem=1` flag into CMake to make it generate an XCode project for Etterna.
I'd be shocked if that had anything to do with anything, but worth mentioning I guess.

The main issue seems to be that \<vector\> (and virtually every other STL header) eventually transitively pulls in \<version\> when building, which includes the garbage files.

bluebandit21:
This change was made already in #1061 , making this PR irrelevant.

PR #1024, opened 2021-08-02
Optimize graphics

staticssleever668:
This also strips metadata from the images. Similar effort has been made before in #26, so this time overall package size didn't change that much, though it's still a bit smaller. Packaged with zstd the difference is about 200KB (18.9 -> 18.7 MB).

Tool used: optipng.
Flags: `-o7 -zm 1-9 -strip all`.


Ulti-FD:
I'm still in the process of finishing up a PR that does touch some of these files. For the sake of avoiding having to deal with *another* merge conflict, it would be better to hold off on this. More so since I'm adding new images that could probably stand to be optimized themselves.

poco0317:
that is why i did not merge this

PR #1038, opened 2021-08-20
High precision timer frame timing

graemephi:
I'm PRing this separately from the audio timing because it's a bit more opinionated and technically changes existing behaviour that maybe some people prefer somehow.

The current implementation is effectively flooring `now()` to milliseconds, and estimating frame times in microseconds, which has the effect of underestimating the time it has to wait to hit the desired fps.  You can think of it as subtracting a random value between 0 and 1000 microseconds in such a way that the average of the random values makes the measured FPS hit the desired FPS with however much resolution it had (looks like 1ms).

Replacing everything with high precision timers you get the opposite problem: if the current time is specified in high accuracy then waiting for the entire frame time to elapse means we definitely wait too long. I haven't thought too deeply about it but I'm pretty sure the amount of overshoot is machine and workload dependent so you can't just slap -250 microseconds on there and call it done. To hit the desired frame time you have to measure the frame times you are actually getting and compensate. The implementation I did was 1) the first thing I thought of 2) not researched in anyway. But it works and is low overhead, so, you know

Now, the opinionated part: v-sync is busted. First, `GameSoundManager::GetFrameTimingAdjustment` was supposed to account for thread scheduling lag by taking advantage of the reference time vsync gives you but didn't check for vsync correctly (that is, it was broken, but even if it wasn't broken, it was still wrong, and why on earth is it in GameSoundManager?). This is fixed.

Next, the current v-sync behavior if a frame limit is set is to _always busy wait before presenting a frame_. This is bad in one probably common case: setting v-sync on and the frame limit to the monitor refresh rate. All this can do is add a frame of lag.

My guess is this happened because the interaction between frame timing, v-sync, and predictive v-sync (also busted) is not obvious, so I moved all frame timing logic off the non-predictive v-sync and framelimit=0 paths. This means normal v-sync ignores any framelimit settings. Predictive v-sync I left alone.

Both of these things can change how the game feels for people who use vsync. imo for the better, but, eh

### DIFF
--- a/src/Etterna/Screen/Gameplay/ScreenGameplay.cpp
+++ b/src/Etterna/Screen/Gameplay/ScreenGameplay.cpp
@@ -778,11 +778,11 @@ ScreenGameplay::UpdateSongPosition(float fDeltaTime)
 		return;
 	}
 
-	RageTimer tm;
+	RageTimer tm = RageZeroTimer;
 	const auto fSeconds = m_pSoundMusic->GetPositionSeconds(nullptr, &tm);
 	const auto fAdjust = SOUND->GetFrameTimingAdjustment(fDeltaTime);
 	GAMESTATE->UpdateSongPosition(
-	  fSeconds + fAdjust, GAMESTATE->m_pCurSong->m_SongTiming, tm + fAdjust);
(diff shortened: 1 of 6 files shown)


poco0317:
superceded by #1051

PR #1051, opened 2021-09-08
Butter

graemephi:
This PR includes the changes in my frame timing PR because I change the RageTimer implementation, which screws up the old RageTimer-based frame pacing. (I didn't think too hard about the ragetimer changes cause I know james wants to junk it eventually.)

Brief history: You can't really trust RageSound's `GetPositionSeconds` on (afaict) anything other than WaveOut. Shakesoda first noticed this with pulseaudio here bfba3a6fe8b3d3d9209e0c4b4b75fe7520d450f5 and xwidghet on directsound later here 427f9f0ff2648a4d2281995f814cde331f66e372

Those changes appear to fix the issue at high frame rates because they reduce the jitter to the reciprocal of your frame rate. (But even, the error scales linearly with audio rate). Since etterna makes use of `steady_clock` now it's pretty straight forward to clean up the resulting jitter, see the comments in `RageSound` for an explanation.

I moved shakesoda's old fix from `GameState` to `RageSound` (and appended my new fix there) because it turns out that every time `GetPositionSeconds` was called, the calling code was doing at least one thing wrong. That's including my code in the previous PR. So, it is no longer the callers responsibility to do anything clever with the value from `GetPositionSeconds` because nobody is clever.

Anyway since I was looking so closely at this I thought I'd go through all the input timing code to see if I could turn up anything and I did 0803206e6d014138cf9773993119b59246e361f1. It's extremely minor. But: someone [_actually went through the trouble_](https://github.com/etternagame/etterna/blob/master/src/arch/Sound/RageSoundDriver_Generic_Software.cpp#L546) however many decades ago of making sure the timestamp you get from soundman does actually correspond to the audio position at that time. This finishes what they started. Combined with my other changes to `RageSound` this means every hit judgement is computed using exactly two calls to `steady_clock::now()`, instead of the up to 5 before.

Since both of these affect how hits are judged I've isolated them in two commits you can easily read and not just take my word for it that I've tested it thoroughly (I have tested it thoroughly.) 

This incidentally fixes vsync

### DIFF
--- a/src/Etterna/Actor/Base/Actor.cpp
+++ b/src/Etterna/Actor/Base/Actor.cpp
@@ -900,7 +900,7 @@ Actor::UpdateInternal(float delta_time)
 			break;
 		case CLOCK_TIMER_GLOBAL:
 			generic_global_timer_update(
-			  static_cast<float>(RageTimer::GetUsecsSinceStart()),
+			  RageTimer::GetTimeSinceStart(),
 			  m_fEffectDelta,
 			  m_fSecsIntoEffect);
 			break;
--- a/src/Etterna/Actor/Gameplay/Player.cpp
+++ b/src/Etterna/Actor/Gameplay/Player.cpp
@@ -1928,25 +1928,15 @@ Player::Step(int col,
 			 bool bRelease,
 			 float padStickSeconds)
 {
-	// Do everything that depends on a timer here;
-	// set your breakpoints somewhere after this block.
-	std::chrono::duration<float> stepDelta =
-	  std::chrono::steady_clock::now() - tm;
-	auto stepAgo = stepDelta.count() - padStickSeconds;
-
-	const auto fLastBeatUpdate = GAMESTATE->m_Position.m_LastBeatUpdate.Ago();
-	const auto fPositionSeconds =
-	  GAMESTATE->m_Position.m_fMusicSeconds - stepAgo;
-	const auto fTimeSinceStep = stepAgo;
-
-	// input data
-	//m_pPlayerStageStats->InputData.emplace_back(!bRelease, col, fPositionSeconds);
-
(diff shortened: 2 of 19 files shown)


src/Etterna/Actor/Base/Actor.cpp
```diff
@@ -900,7 +900,7 @@ Actor::UpdateInternal(float delta_time)
 			break;
 		case CLOCK_TIMER_GLOBAL:
 			generic_global_timer_update(
-			  static_cast<float>(RageTimer::GetUsecsSinceStart()),
+			  RageTimer::GetTimeSinceStart(),
```
graemephi, line 903: This was a bug right. It looks like a bug, It used to be seconds before it somehow become microseconds

src/Etterna/Actor/Gameplay/Player.cpp
```diff
@@ -1928,25 +1928,15 @@ Player::Step(int col,
...
-	}
+	const auto fMusicRate =
+		GAMESTATE->m_SongOptions.GetCurrent().m_fMusicRate;
+	const auto fHitUpdateDelta =
+		(GAMESTATE->m_Position.m_LastBeatUpdate - RageTimer(tm)) - padStickSeconds;
+	const auto fMusicSeconds =
+		GAMESTATE->m_Position.m_fMusicSeconds - (fHitUpdateDelta * fMusicRate);
+	const auto fSongBeat =
+		GAMESTATE->m_pCurSteps ? m_Timing->GetBeatFromElapsedTime(fMusicSeconds)
+							   : GAMESTATE->m_Position.m_fSongBeat;
```
graemephi, line 1939: Two things here: padStickSeconds gets scaled with rate which seems wrong, and `m_pCurSteps` is nullchecked before accessing `m_Timing`. I don't understand either of these so I just reproduced the current behaviour

poco0317:
supercedes #1038

PR #1059, opened 2021-09-28
Rename flashWindow to requestUserAttention, and implement for macOS

bluebandit21:
flashWindow doesn't actually flash the window on macOS.
requestUserAttention seems like a nicer platform-agnostic name for the function


bluebandit21:
Note: I think Clang-Format got called on PlatformWin.cpp for the first time in a long time, if ever.
This creates a bizarrely large diff for what should be a simple PR.

I can specifically disable clang-format for this commit and re-create the PR if desired.

PR #1066, opened 2021-10-11
Fix macOS SetupFFMPEG Cmake issue which lead to build failing

bluebandit21:
The FFMPEG Setup Cmake deletes the file VERSION from the ffmpeg generated folder, as it conflicts with the C++20 <version> header.

However, the cmake uses rm without flags, which errors out if the file has already been deleted by a previously succesful run of the cmake.
Adding the "-f" flag makes it proceed even if the file does not exist to be deleted.


jameskr97:
👍

PR #1071, opened 2021-10-21
⌛C++20 Upgrade

jameskr97:
# Upgrade to C++20

## Must be merged after glfw

It's time for an upgrade. The only requirement changed is that we must use at minimum:
- Windows: MSVC 19.20 (Visual Studio 2019)
- Mac: Xcode 11.0
- Linux: gcc-8, clang-9

Active Developers: Please test locally to ensure you can still compile.


nico-abram:
Can we remove extern/fmt and use std::fmt if we're using C++20?

nico-abram:
I tested this locally and it built fine with VS2022 and playing a song worked fine. Like you said on discord, it doesn't seem to be necessary to merge this after glfw. LGTM

PR #1078, opened 2021-11-09
Add alpine dependencies to Building.md

nico-abram:
I built the game on alpine earlier and these were the dependency names I needed for it's package manager (apk). Idk if it makes sense to add a (presumably) niche usecase like that to the instructions


Ulti-FD:
There's 4 other spots with dependencies that you didn't add yet

nico-abram:
Thanks @Ulti-FD, good catch. Added them. The only weird one is the ninja command, It seems ninja isn't available in any of the alpine apk repos (main, community, testing or edge) but a ninja clone called [samurai](https://github.com/michaelforney/samurai) is, which is what I used and it built fine.

PR #1079, opened 2021-11-14
Fix bug poco found

nico-abram:
When a DifficultyList had more than LINES_INITIAL_ALLOCATION lines we crashed

See the comment in the added line, I tested this on scwh with the "BUTTERFLY" simfile poco shared on discord and it didn't crash (It crashed before the change).

cc @poco0317


poco0317:
replaced by #1080

PR #1083, opened 2021-11-30
Default beat skin upgrade

Jolomann:
Let's see if I remember how to do a PR

Mainly takes care of bringing the skin to a usable state and removing ties to stolen elements like the notes and tap explosions

Before and after comparison here https://www.youtube.com/watch?v=Bp0RlJqgs8o

- New explosions
- New notes and holds (including missing red hold)
- There are no longer gaps between notes and receptors
- Removed receptor flash
- Removed receptor shrink on press
- New column lasers (that are also used now, which they weren't previously)
- New mine
- Removed unused elements
- Reduced duplicate elements and compressed images (roughly 400KB -> 60KB)
- Fixed fallback elements showing during gameplay lead-in
- Tried to hide things in Noteskin.lua but the ret.Blank table had no effect. Added "zBlank.png" and redirected to that for blanking instead


staticssleever668:
IMO something should be done about these new note textures, they are blurry and ugly. I also don't think explosion texture should be changed, new one doesn't bring anything useful, it's just different. Current one if I remember correctly is the default in Lunatic Rave 2, so I think it fits better.

Not up to me to decide in any way, just my opinion. ;)

Jolomann:
I think you raise some good points. Normally I would keep things in high res but for this I specifically kept things low res to not bloat things over the original, and with the softer gradients it looks very blurry compared to the original. I can play around with the textures a bit to get something less "muddy". I wasn't too critical about them because all I was looking for was something that was free to use aka not literally stolen.

As for the explosions they're getting changed to make sure they're not infringing on any copyright. Them being different is just for the sake of not making direct knockoffs. The previous ones are literally from beatmania iidx, not default lr2. Even if they were from lr2 I don't believe that would fit under etterna's license, but I could be wrong about that. There was also no hold explosion, which this adds

Jolomann:
The first commit was missing more than half the noteskin which is now fixed

Adjusted the notes @staticssleever668 , much better this way, thanks

poco0317:
this is merged but git didnt believe me

ecafree2:
image
i found that it's a bit off to the center, maybe it can be adjusted from the main program i think

Jolomann:
Will look into the alignment soon™️, thanks for pointing it out

Jolomann:
@ecafree2 Notes were overlapping which is fixed here https://github.com/etternagame/etterna/pull/1138

The overall alignment could also be fixed, but I think it's best to leave janky trickery out of a default skin. If possible, it would be better to have Key4 be in screen centre instead of Key3, but that's best to do with the main program like you mentioned

PR #1119, opened 2022-01-09
Ver0.71.1 Chinese translation for TilDeath, _fallback, and Rebirth

SharpnelXu:
Tested with newest 0.71.1 executable


poco0317:
Wasnt expecting this so quickly. I planned on adding full translation support to Rebirth at some point and I guess it needs to be done now. Some strings are a little non-trivial to have translated for now

SharpnelXu:
> Wasnt expecting this so quickly. I planned on adding full translation support to Rebirth at some point and I guess it needs to be done now. Some strings are a little non-trivial to have translated for now

In this case I can hold back from translating Rebirth. I don't use Rebirth so I don't have a good idea of what everything is doing.

poco0317:
ok this literally took forever for me to get to but I've finished translation support for Rebirth on the latest develop as of now. go ahead and redo/rebase this PR on that if you want

SharpnelXu:
Finally get back to this, not sure I if did this correctly tbh

poco0317:
after switching the target branch to develop, its looking good. i can tell it isnt finished, but it would work in its current state

poco0317:
gonna merge this so we can have a good point to pick up and continue these changes from

PR #1138, opened 2022-02-25
Fix misaligned notes in default beat skin

Jolomann:
Just fixing a small oversight. White/gray notes were overlapping into blue columns, but now they don't. Also updated white laser to be consistent with other textures (no res tag)


poco0317:
please for the love of god reset your develop to the latest develop

PR #1142, opened 2022-03-07
Update Contributing.md

zoey-on-github:
put a space between "is" and "a" in the lua section.


nico-abram:
LGTM
Although I think this should probably go against the develop branch instead of master

PR #1154, opened 2022-04-07
Update Contributing.md

zoey-on-github:
removed the part that says "I've made more or less a working package for debian" as the link seems dead, as it just leads to a 404 page now.


nico-abram:
Oops, I probably deleted that when cleaning the branches on my fork. Thanks for noticing that. If you were looking for it, I *think* the debian-pkg branch on the main repo was more or less what was there? This one https://github.com/etternagame/etterna/tree/debian-pkg , just look at the last commit. But beware that it's (1) super old and probably outdated, and (2) I know nothing about packaging so it's likely doubly wrong.

PR #1171, opened 2022-05-14
Suppress a whole lot of EnumTraits warnings

bluebandit21:
These warnings are only indicating that a linker error might occur later ... which it doesn't.

The specific warnings are caused by us instantiating the various EnumTraits static member variables in a different translation unit than the one building Check() and friends.

However, as we *do* instantiate them, the warning isn't helpful and just spits out a lot of noise into our build output.

### DIFF
--- a/src/Etterna/Models/Misc/EnumHelper.h
+++ b/src/Etterna/Models/Misc/EnumHelper.h
@@ -106,7 +106,7 @@ EnumToString(int iVal,
 	}                                                                          \
 	}
 
-/* Disable errors about instantiations of static member variables of the
+/* Disable warnings about instantiations of static member variables of the
  * templated class not being defined. These are all defined in different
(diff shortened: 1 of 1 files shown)


poco0317:
cleanest linux gha run ive ever seen by literally tens of thousands of lines

nico-abram:
I agree with silencing this. I looked at the motivation for the warning in LLVM (https://reviews.llvm.org/D16396 / https://reviews.llvm.org/D12326#240286) and they said 

> For valid code, this means we'd warn on templates/temploids which have their definition and all relevant explicit instantiations tucked away in some source file (but for which no explicit instantiation declarations are provided in the relevant header file) -- uncommon, but an idiom used by some. There is at least a syntactic solution we can suggest for people who see the warning: add the missing explicit instantiation declarations to the header file.

Which sounds like exactly this scenario.

PR #1182, opened 2022-05-29
The Rebirth theme is the default, not Til' Death.

Plextora:
In this commit, I fixed mentions of Til' Death being the default. The Rebirth theme is now the default for Etterna, not Til' Death. I capitalized the "e" in Etterna in the path to the en.ini file and put it inside a `` block, and added a space before the parentheses on the last sentence of the paragraph.

### DIFF
--- a/Docs/Contributing.md
+++ b/Docs/Contributing.md
@@ -38,7 +38,7 @@ The project is sorely lacking in documentation. The current aim is to document t
 
 ## Translating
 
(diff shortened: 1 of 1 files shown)


nico-abram:
The github repository is called "etterna" without a capital E, so I think the path was fine before.

Plextora:
Thanks for pointing that out, I should've thought of that!

poco0317:
replaced by 68eee7b2bac82f62dde9576f46c4778500b57673  5bad00d98b22307a288620791a148ebd22e0b966

PR #1191, opened 2022-06-25
Allow for use of new XCode Build System without errors

bluebandit21:
The new XCode build system doesn't allow the same file to be generated by two independent targets without
 having it be a dependency of some common ancestor first.

This could be worked around by using the old XCode build system with a special CMake flag, but this produces
 some rather spooky warnings about the fact that it will be removed at some indeterminate point in the future.

Both the cURL and LuaJIT CMakes violate this requirement.

This fixes both very simply --
For cURL, we add in a third random target to be a parent to both that generates the common files, fixing the issue.
For LuaJIT, we do the same, although the parent target already exists (containing only the shared file
 complained about in the first place)


nico-abram:
To my non-CMake-expert eyes this looks fine. The curl stuff is just for pdf/man pages it seems? And the luajit thing seems to be about this lj_vm_s target : https://github.com/etternagame/etterna/blob/5f2c6278a5824a60af9f8ca03e468acf745e9e4a/extern/luajit/src/CMakeLists.txt#L367-L375
which only depends on a custom command (which I believe executes luajit/src/host/buildvm.c to compile the luajit/src/vm_ARCH.dasc files into object files)
"luajit_sources" contains on that output object file too because of this line https://github.com/etternagame/etterna/blob/5f2c6278a5824a60af9f8ca03e468acf745e9e4a/extern/luajit/src/CMakeLists.txt#L488
And libluajit depends on "luajit_sources" https://github.com/etternagame/etterna/blob/5f2c6278a5824a60af9f8ca03e468acf745e9e4a/extern/luajit/src/CMakeLists.txt#L499
So your fix of making libluajit depend on the lj_gen_vm_s target seems fine to me, I don't see how that could go wrong.

@jameskr97 any thoughts on this PR?

nico-abram (approved):

jameskr97:
One month late, but LGTM.

PR #1199, opened 2022-09-10
Slightly cleanup DLMAN

nico-abram:
commit messages have the description of the stuff I did. Aggregated:

 - Removed unused ReadThis class and functions
 - removed unused cpuid in ComputerIdentity()
 - Inlined progressfunc and write_data definitions to the only place
   they're used as capture-less lambas
 - Inlined checkProtocol function which had a confusing name and was
   only used once (Also changed default to https)
 - Removed the hacky EncodeSpaces function, replaced it with curl's url
   encoding function (curl_easy_escape) in the places where I think it
   was needed (tested score leaderboard requests and pack downloading
   and they seem to work alright)
 - [Remove ends_with function from DLMAN (Already in RageUtil.h)](https://github.com/etternagame/etterna/commit/5b91178ce31e88647ccdde8320e24007d8fb9371)
 - [Remove unused addFileToForm function from DLMAN](https://github.com/etternagame/etterna/commit/d044791f67b71f535629fece23cd5086fd462f69) (unused)
 - Remove SetCURLPostToURL which was only called in 2 places and only
added one line over SetCURLURL
 - Remove an unused CURL handle fromDownloadManager::RefreshTop25 which I think we were leaking.
 - Move CURL handle deletion of the Download struct from ~DownloadManager to it's destructor ~Download
 - Removed the unused Done() callback from Download
 - Remove split (unused, probably replaced with splitPath at some point
   and not removed)
 - Renive findStringIC (unused)
 - Remove duplicated GetChartLeaderboard AddMethod call
 - Replace a couple 3-line empty lambda definitions which just add noise
   with {} literals which are equivalent
 - [Remove unused CURLMsg* param from DLMAN callbacks](https://github.com/etternagame/etterna/pull/1199/commits/82ad755650a374aaf6b80e5621583d9c1c3e60d9)

### DIFF
--- a/src/Etterna/Singletons/DownloadManager.cpp
+++ b/src/Etterna/Singletons/DownloadManager.cpp
@@ -442,13 +442,13 @@ DownloadManager::UpdateHTTP(float fDeltaSeconds)
 		for (size_t i = 0; i < HTTPRequests.size(); ++i) {
 			if (msg->easy_handle == HTTPRequests[i]->handle) {
 				if (msg->data.result == CURLE_UNSUPPORTED_PROTOCOL) {
-					HTTPRequests[i]->Failed(*(HTTPRequests[i]), msg);
+					HTTPRequests[i]->Failed(*(HTTPRequests[i]));
 					Locator::getLogger()->warn(
 					  "CURL UNSUPPORTED PROTOCOL (Probably https)");
 				} else if (msg->msg == CURLMSG_DONE) {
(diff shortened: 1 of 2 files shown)


nico-abram:
Note: The alias I removed had a lowercase `b` instead and *is* used in 'til death, so I added it back

nico-abram:
Superceded by https://github.com/etternagame/etterna/pull/1200

PR #1202, opened 2022-10-09
Some calc floating point stability fixes 

graemephi:
The calc is unusually sensitive to tiny changes in row times. For example depending on whether they are computed [here](https://github.com/etternagame/etterna/blob/877d46ae88f9d7d6c887a7e954192bd757399cc1/src/Etterna/Models/Misc/TimingData.cpp#L1400) or [here](https://github.com/etternagame/etterna/blob/877d46ae88f9d7d6c887a7e954192bd757399cc1/src/Etterna/Models/Misc/TimingData.cpp#L1418) can cause a large change in the result, even when they are equivalent (i.e., bpm change after the last non-empty row)

This is kinda academic but it seemed like an interesting thing to track down and fix so here we are. This PR moves most files around randomly in a range of around 0.5 MSD. It is apparently a bit better on average but that may just be chance

It turns out there were 2 sources of instability:
 - Comparisons of row time deltas. This one is where most the instability was coming from. Basically for row times `t0`, `t1`, `t2`, `t3`, all a 16th apart, you don't necessarily have `(t1 - t0) == (t3 - t2)` due to rounding error. This matters for the couple places where the calc branches on one being greater than the other. For equal deltas, sometimes you get lucky and it's less or equal, sometimes you get unlucky and it's greater.
 - Inconsistent interval binning on certain BPMs. Since the first non-empty row is set to time 0 and intervals are half a second, any bpm which lines up rows onto half seconds has many rows that lie exactly on interval boundaries. Those are representable in floating point but may be missed due to rounding error. If it comes under, rows get binned into the previous interval. So e.g. the calc may see oscillating NPS density where it should be constant. Smoothing and sequencing makes this less bad but it's still weird

Aside from the changes in this PR, another thing that would help is rounding the row times (after rate scaling) to a power of two, like `1/1024`. That'll fix interval binning and it'll make row time deltas exact. But it will only fix comparisons of the form `(t1 - t0) == (t3 - t2)`, if you do anything more complicated you end up in the same situation. So I didn't bother

### DIFF
--- a/src/Etterna/MinaCalc/CalcWindow.h
+++ b/src/Etterna/MinaCalc/CalcWindow.h
@@ -182,7 +182,7 @@ struct CalcMovingWindow
 		// we can basically just branch to ccacc or acca checks depending on
 		// which value is higher
 		bool o;
-		if (_itv_vals[4] > _itv_vals[5]) {
+		if (any_ms_is_greater(_itv_vals[4], _itv_vals[5])) {
 			// if middle is higher, run the ccacc check that will divide it
 			o = ccacc_timing_check(factor, threshold);
 		} else {
--- a/src/Etterna/MinaCalc/Dependent/HD_PatternMods/Chaos.h
+++ b/src/Etterna/MinaCalc/Dependent/HD_PatternMods/Chaos.h
@@ -49,7 +49,7 @@ struct ChaosMod
 		// previous value
 		const float b = ms_any.get_last();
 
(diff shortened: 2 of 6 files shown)


poco0317:
this is (maybe) kind of why we considered using chrono durations instead of floats for all the second and millisecond values. but when i tried this it didnt immediately work for all the copy pasting effort so i gave up and by that point mina resigned to not caring enough to do it himself

good enough

PR #1216, opened 2022-12-11
Fix websocketpp to be compliant with c++20

g-robertson:
websocketpp dependency uses template specialization on class constructors which isn't allowed in c++20, this change fixes that


poco0317:
related https://github.com/zaphoyd/websocketpp/pull/1060

poco0317:
i will merge this if you first update websocketpp to the latest upstream version 0.8.2, as we appear to be using 0.8.1

PR #1221, opened 2022-12-26
(Again) disable use of libssh2 when building cURL

bluebandit21:
Without this change, we run into the problem where our GHA builder has the libssh2 library installed (and so the built cURL dynamically links to it) but the average mac user does not, leading to the application immediately crashing on start.

This was already fixed once before, but with the recent(ish) upgrade to a newer version of cURL than what we'd been using in f1dd6fe, the option CMAKE_USE_LIBSSH2 was renamed to CURL_USE_LIBSSH2, which silently made our forcing of that setting to OFF no longer work.

### DIFF
--- a/extern/CMakeLists.txt
+++ b/extern/CMakeLists.txt
@@ -13,7 +13,7 @@ set(BUILD_SHARED_LIBS OFF) # Tell CMake to prefer static libs
 
 # External Libraries
 ## Curl Specific Options
-set(CMAKE_USE_LIBSSH2 OFF CACHE BOOL "" FORCE) # Disable curl libssh2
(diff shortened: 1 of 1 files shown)


jameskr97:
I will review this tonight.

nico-abram:
Looks reasonable to me. For anyone else reviewing, the cmake option is used/defined here

https://github.com/etternagame/etterna/blob/f07cd674bedf3730da2fa3d9f1c79349cc9785d7/extern/curl/CMakeLists.txt#L746

https://github.com/etternagame/etterna/blob/f07cd674bedf3730da2fa3d9f1c79349cc9785d7/extern/curl/CMakeLists.txt#L752-L768

Issue #1215, opened 2022-12-11
[Feature Request]: Work around openssl1.1 being required but not easily available on modern Ubuntu derivatives


bluebandit21:
### Is there an existing issue for the feature?

- [X] I have searched the existing feature requests

### Describe the Feature

Etterna dynamically links in libssl1.1 as a runtime dependency.

Unfortunately, as of Ubuntu 22.04, libssl1.1 is no longer easily available as they have fully deprecated it in favor of libssl3.

As a result, modern Ubuntu and all derivatives require something not easily obtainable even for users using our prebuilt Linux binary.

We should handle this in some fashion or other, either
a) Including instructions for acquiring libssl1.1 on modern Ubuntu derivatives
b) Trying to fix the root problem, that we require a now-ancient libssl version (can we statically link it in, or upgrade Etterna to use libssl3 instead?)

### How Does The Feature Add To The Game?

Currently, users using modern Ubuntu/Ubuntu derivatives are unable to easily build Etterna or even use our pre-built binary without having to obtain a runtime dependency that now requires some technical know-how to obtain.

This problem makes it much harder for new Linux players to be able to play the game.

poco0317:
based on this https://cmake.org/cmake/help/latest/module/FindOpenSSL.html

we could probably move required cmake version to 3.19 and change this line with no other issues?
https://github.com/etternagame/etterna/blob/23bb8e90634183f998fea7bba83689e26eed9abb/CMakeLists.txt#L71

bluebandit21:
An additional point of concern mentioned by #1326, is that OpenSSL1.1 will now no longer receive any security updates, as per https://openssl-corporation.org/post/2023-09-11-eol-111/

That means we should really be working on switching to OpenSSL3.x rather than trying to improve the ease of users installing OpenSSL1.1 after its deprecation.

poco0317:
superceded by https://github.com/etternagame/etterna/issues/1330

PR #1229, opened 2023-01-10
Make OpenSSL static link on all operating systems preferentially

g-robertson:
This should resolve issue #1215 as it will make the binaries generated for Linux static link OpenSSL so ubuntu 22.04 should not need to have OpenSSL 1.1 to function.

The reason this was chosen over switching to OpenSSL 3.0 with maintaining dynamic linking on Linux is because doing this would create the same problem for Ubuntu 20.04 LTS as they are stuck with OpenSSL 1.1. The CI for Etterna also uses Ubuntu 20.04 which means trying to switch to OpenSSL 3.0 would make that build fail without manually grabbing the package from the Ubuntu 22.04 repositories.

The reason dynamic linking is still allowed in the build process is for operating systems (Arch Linux) that lack static versions of packages and must dynamically link if they want to compile the game (see #952)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,10 +63,16 @@ target_include_directories(Etterna PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/src)      
 ## Package Includes
 ### OpenSSL is not used directly by our program, but we have to use OpenSSL first
 ### in order to statically link. Once find_package is run once, it's results are cached
-set(OPENSSL_USE_STATIC_LIBS ON CACHE BOOL "" FORCE)
+set(OPENSSL_USE_STATIC_LIBS ON CACHE BOOL "")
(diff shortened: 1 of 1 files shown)


nico-abram:
Is there a reason to automatically switch to dynamic linking in the cmake script if we fail to find a static openssl? That seems bad to me if the user explicitly sets  `OPENSSL_USE_STATIC_LIBS `

g-robertson:
The aim was for compilation to "just work" without any additional flags on the same devices that it currently does regardless of if they don't have dynamic or static versions of openssl. I could make it error if it's not found statically with an error message more explicit in what needs to be done in order to make compilation work on systems without static versions of OpenSSL (e.g. "No static version of OpenSSL was found, try setting the flag `-OPENSSL_USE_STATIC_LIBS=OFF`".)

PR #1236, opened 2023-01-21
Add and implement Core::setClipboardText

bluebandit21:
Currently implemented for macOS
Implementations are stubbed out for Unix and Windows Core platform

### DIFF
--- a/src/Core/Platform/Platform.cpp
+++ b/src/Core/Platform/Platform.cpp
@@ -63,6 +63,11 @@ int getClipboard(lua_State* L) {
     return 1;
 }
 
+int setClipboardText(lua_State* L) {
+	lua_pushboolean(L, Core::Platform::setClipboardText(SArg(1)));
+	return 1;
+}
+
 int setCursorVisible(lua_State* L){
     Core::Platform::setCursorVisible(BArg(1));
     return 0;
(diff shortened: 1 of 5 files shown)


jameskr97:
Will complete linux complete impl.

poco0317:
Added and tested windows implementation
Added lua binding

poco0317:
Tried for a few hours to do the linux implementation. Cant get it to work in ubuntu 20.xx.
Here's what I've tried: https://github.com/etternagame/etterna/compare/a0bad9d791e44b4172c2007c1ef51f83a6dba0c5...poco0317:etterna:fa52f2f4aa97a28a13d78dbd6873120ea9344306?expand=1

Basically its a horrible headache to interact with linux because theres multiple window implementations and we assume X11 is good enough but it probably isnt. GLFW or something would probably solve this but hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh

More resources:
https://www.uninformativ.de/blog/postings/2017-04-02/0/POSTING-en.html
https://tronche.com/gui/x/xlib/event-handling/manipulating-event-queue/
https://tronche.com/gui/x/xlib/window-information/
https://handmade.network/forums/articles/t/8544-implementing_copy_paste_in_x11
Consider XPeekEvent vs XNextEvent (but dont sit in a loop forever and break the application)

PR #1237, opened 2023-01-21
Modify Core::Platform::getLanguage to work with older macOS versions

bluebandit21:
Locale.NSLocale.languageCode was only added in macOS 10.12. However, the same information can be obtained a slightly different way that is non-deprecated and works all the way back to macOS 10.4.

This small change makes Etterna compatible with (at least) El Capitan again excepting the known issue with OpenSSL on High Sierra and earlier users.

### DIFF
--- a/src/Core/Platform/PlatformMac.mm
+++ b/src/Core/Platform/PlatformMac.mm
@@ -112,7 +112,7 @@ namespace Core::Platform {
     }
 
     std::string getLanguage(){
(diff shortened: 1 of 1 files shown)


bluebandit21:
Note (from Hidden):
Although 10.10 is the earliest version we theoretically support (based on our root CMakeLists.txt), without this change we only actually support back to 10.12 so it's still relevant.

PR #1249, opened 2023-05-18
til death fix updating of judgement bar fill, percent, count, and left/right cbs

Darkmans1660:
f846fc05ea343fbaa380ace932236e16f9f8a5cb

Before these changes, The fill, percent, and count would use the wrong judge when selecting a score from the right-side leaderboard.

For example, if a score's judge is 5, and I change the judge on the evaluation screen to 6, then click on the same or a new score, the fill, count, and percent update using the score's judge instead of the set judge (6).

After these changes, this is no longer the case since the fill, count, and percent now use the judge you have selected.

ef3abb618d18f567504d29de77f2d0bc4086ec62

Before these changes, the left and right cb counts update using the score you entered the evaluation screen with when changing judges instead of the selected score, causing inaccurate values if you select a different score and change judges.

After these changes, the updates use the selected score instead.

### DIFF
--- a/Themes/Til Death/BGAnimations/ScreenEvaluation decorations/default.lua
+++ b/Themes/Til Death/BGAnimations/ScreenEvaluation decorations/default.lua
@@ -715,7 +715,7 @@ local function scoreBoard(pn, position)
 					self:zoomx(frameWidth * newjudgecount / totalTaps)
 				end,
 				ScoreChangedMessageCommand = function(self)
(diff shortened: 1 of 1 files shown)


Darkmans1660:
Sorry if I did anything wrong with this pull request. I haven't made one in a long time.

I also don't have much experience with Lua or etterna theming, so hoping I didn't do anything wrong.

poco0317:
this is almost correct. or maybe it breaks something. now when i turn on custom windows and then click a score, the judgments for the play are incorrect

PR #1255, opened 2023-05-23
Added new SortMode "StepAuthor" for Issue#1244

SaucySensation:
Added new SortMode "StepAuthor" for Issue#1244. This sort mode uses the title as a secondary metric to sort under.


poco0317:
i renamed StepAuthor to Author to match rebirth, and reverted the song.cpp/song.h change because this was feeding garbage data to the lua hooks

Issue #1263, opened 2023-07-12
[Feature Request]: Re-add right click to close various ui elements a la td


MinaciousGrace:
### Is there an existing issue for the feature?

- [X] I have searched the existing feature requests

### Describe the Feature

Right click to cancel/close screens/panels such as
- help screen
- download panel (when over panel frame)
- etc

also right click to revert to general tab when on other tabs while over tab frames

### How Does The Feature Add To The Game?

should slightly improve ui navigation speed

PR #1276, opened 2023-09-06
Right click to close elements on rebirth

pedrosoutinho:
#1263 

Added right click to close the following elements:

- Profile assets
- Search
- Downloads
- Help
- Settings


poco0317:
i cant accept these because the right click functionality as is overlaps with right click to pause. the right clicking has to be gated to a particular area for things such as asset settings

pedrosoutinho:
I'll look into that

poco0317:
for this type of thing basically what i did was check isOver(bg) if the thing has a background or create an invisible quad that acts as the click area but with very particular layering so it doesnt break other overlapping buttons

PR #1277, opened 2023-09-12
Add filter preset functionality

### LINKED ISSUE #1145: [Feature Request]: Custom filter presets
### Is there an existing issue for the feature?

- [X] I have searched the existing feature requests

### Describe the Feature

Etterna has a comprehensive song filter search system, which people often use to create a detailed search. However, due to the nature of the game, people who utilize the feature will be putting in the same several searches over and over when they are focusing different things within a session (e.g. warmups for specific skillsets, pushing for skillsets, ranges for sightread accuracy). 

So it was proposed to create a **preset system** for filter settings, which effectively means that the user would be able to more easily access commonly used searches with minimal effort after it is created. Upon hearing this and considering it an interesting idea, I took it upon myself to create a mockup demoing how such a feature could be included within the current search tab, and hopefully also be a reasonably intuitive system. Below are several mock-ups, how things should function will be outlined with each image. 

### First opening the filter tab
1 - Filter no action
- When a user first sees the filter tab, there are a few new settings viewable: **Preset, Save, Delete,** and **Rename**
- When no preset is selected, **Delete** and **Rename** are grayed out and unusable 
- When no filter settings have been modified, **Save** is grayed out and unusable, it enables once anything has been changed
- If no filter is currently active, the tab always defaults to **None** for the preset
- If the user has saved presets, they can either click the arrow keys, or use the scroll wheel to cycle through existing presets
- Upon hover, the white box appears behind the setting text like within the options menu
- When a filter preset is cycled to, all the filter settings change to reflect the preset (user still **must hit apply to search it**) 

### Making a new filter preset
2 - Filter created with no preset
So now that the more general rules are laid out, let's consider the user making a preset of their own. In this case, I've decided I want to make an advanced level JS filter. Now that I have tinkered with filter settings, the **Save** option is now white and will allow me to interact with it! Note that like other elements in this tab, it should darken a little on hover. 

### Saving my preset
3 - Filter created and save initiated
Now that I am ready to save my special preset, I clicked the **Save** button. This opened a **dialogue box** which prompts me to give my preset a name. I can press escape or the exit button to cancel this action, or I can type the name, press enter, and have a brand new preset in my client. When it saves, it should set your preset to that new one, regardless of how many the user currently has. 

If the user has multiple presets, they should be **sorted alphabetically** so they can be easily sorted through. 

### My new preset
4 - Save complete and hovered on setting
The preset is saved and ready to go! Now that I am actually hovered on a preset, note that more options are available to the user. 
- When on a specific preset, **Save** becomes **Update.** This will allow players to make adjustments to their preset and bypass the dialogue box as a name and file already exist. If a user wants to create a new preset, they should go back onto **None**
- **Delete** and **Rename** are now accessible. If **delete** is clicked, the preset will be removed and the user will move to the next filter preset in the cycle (basically as if the user clicked the right arrow once). 

### Renaming my preset
5 - Rename a preset
So let's say I made a mistake while saving my preset, or want to change the function of it a bit, I need a way to alter the name. When the user clicks on **rename** it will bring up the **dialogue box** effectively like saving, but it updates the name of that file instead of creating a new preset. Like with saving, this can be cancelled with escape or by pressing exit. 
- Like with other custom named settings, the text if too long will be squished. 
- The right triangle moves with text length, but will have a maximum right distance due to this squishing, obviously still needing to stay on screen. 

### Other minor notes
- **Reset** and **Apply** are now on the same line, instead of occupying two lines
- All first words on the right side of the tab are left aligned together
- **Apply**, the **left triangle**, and **Delete** are all left aligned with the **filter matches number**

Other ideas were also proposed and hopefully will be included within the comments of this issue. Hopefully this will be enough groundwork to allow for a clear understanding of how this feature could be implemented. 

### How Does The Feature Add To The Game?

While it's explained within the above description, the main thing this adds is QoL for users who currently take advantage of the filter system. Players will have many sessions and often will want to do the same several things, so allowing them to save time within the filter tab for ease of use would be quite beneficial. 

It also allows additional personalization by letting players create and name specific sorts they use often. Ultimately I think this will provide extra incentive to be very particular with a filter so that you can use it again in the future without having to constantly drag everything back into place, especially if you're using multiple filters in a single session. 

### Additional Context

_No response_

Alkorith:
This is a feature patch to mostly resolve #1145. I'm not much a UI/UX guy so I don't know how to go beyond what I have done so far.

# Rebirth Screenshots

## Player pressed "Export"

# Til Death Screenshots

## Player pressed "Export"

# Revisions

Rebase 1: 

https://github.com/etternagame/etterna/compare/1a5e97b8d1fc137e3774d25d7fa95ba83acf26e4..9dae877d4b4f6ee9ff7295a3a43879967ebca459


poco0317:
save and export in this are rather redundant and should probably both go in the same folder. 
i am interested in merging this, but i would prefer to know at least what we want the ui to look like even if it isnt coded yet

Alkorith:
> save and export in this are rather redundant and should probably both go in the same folder - so we would eliminate one of those buttons.

I'd rather keep the ability to have a default preset loaded on startup so I can remove the export button in the next revision. 

> i am interested in merging this, but i would prefer to know at least what we want the ui to look like even if it isnt coded yet

Honestly, I thought just having the button as fine as is. I don't really know how else to improve the UI/UX. If you have any suggestions, please let me know.

Issue #651, opened 2019-11-24
Missing linux installation


Evils-Devils:
Packaging this was requested for [nixpkgs](https://github.com/NixOS/nixpkgs)

**Describe the bug**
The compiled binary does not get installed to the specified CMAKE_INSTALL_BINDIR
It ends up in the downloaded source directory and fails with "cannot find songs" if moved to the bin dir.

**To Reproduce**
https://github.com/Evils-Devils/nixpkgs/blob/etterna/pkgs/games/etterna/default.nix

**Expected behavior**
All executables in CMAKE_INSTALL_BINDIR and able to find their (immutable) assets in CMAKE_INSTALL_PREFIX

**Desktop (please complete the following information):**
 - OS: NixOS
 - Version: nixos-unstable
 - 64bit

**Additional context**
build output etc [here](https://paste.ee/p/yIFyQ)
Nix puts everything in /nix/store and everything there is immutable (read-only)
$HOME is writable

nico-abram:
Thanks for trying to package the game and opening this issue!

What should we do to fix this? Would making our cmake respect CMAKE_INSTALL_BINDIR and copy all the folders the game expects to find in HOME (Themes, RandomMovies, Songs, Screenshots, Save, NoteSkins, Announcers, Assets, BGAnimations, BackgroundEffects, BackgroundTransitions, Characters, Logs) to CMAKE_INSTALL_PREFIX do it? Do we want to put stuff like the license file, /GameTools, /Data and /Docs in the CMAKE_INSTALL_BINDIR ?

Tangentially related: Do we want to change the folder we use under HOME here to etterna? https://github.com/etternagame/etterna/blob/425f10cfd58473754bba9021fc6f441e1accaefc/src/arch/ArchHooks/ArchHooks_Unix.cpp#L423

Also only tangentially related: I think it's likely that putting /Assets in the "non-portable" (Iirc HOME and %appdata%) game folder doesn't work (Haven't tested it) now that I think about it.

Evils-Devils:
as i have never used this application, i'm probably not the best person to advice on a solution

CMAKE_INSTALL_BINDIR, if supplied should definitely be respected and used as a path to install the executable
with nix, CMAKE_INSTALL_PREFIX gets set to a path in /nix/store, and this will become immutable once it's done there
anything that etterna needs and doesn't need to be modified should be placed there
[...]

PR #1282, opened 2023-09-29
Configurable virtual fs mounts based on environment variables

phantamanta44:
This PR is a first step towards addressing #651. It allows users (or package maintainers) to customize the Etterna virtual filesystem mounts using environment variables so that the game installation itself can remain immutable while any variable data is stored elsewhere. This is especially necessary for a NixOS release, as the Nix store is read-only.

More specifically, this PR does the following:
* Causes Etterna to check the env var `ETTERNA_ROOT_DIR` on startup. If present, its value is used for the path of the main mount; otherwise, the current behaviour is used, namely mounting the installation directory.
* Causes Etterna to check the env var `ETTERNA_ADDITIONAL_ROOT_DIRS` on startup. If present, it's parsed as a list of paths that are treated identically to the paths listed in the "AdditionalFolders" setting. The list delimiter is the OS-dependent path separator, namely `;` on Windows and `:` everywhere else.
* Causes Etterna to check the env var `ETTERNA_ADDITIONAL_SONG_DIRS` on startup. If present, it's parsed as a list of paths that are treated identically to the paths listed in the "AdditionalSongFolders" setting. As above, the separator is OS-dependent.
* Adds an `ASSET_DIR` CMake variable that can be configured to define where game assets should be installed. If absent, defaults to the current behaviour, namely to `INSTALL_DIR`. The purpose of this change is to allow assets to be separated from the rest of the game data so that they're easier to subsequently copy elsewhere. Could be useful for distribution on some Linux flavours.
* Makes `INSTALL_DIR` similarly configurable.
* Disables the CMake installation commands for crashpad binaries if Etterna is built without crashpad.

This last change comes up because the crashpad build task attempts to run built-in build tools (e.g. `extern/crashpad/buildtools/linux64/gn`) that fail to run on NixOS because of linking issues. Potentially, a more drastic refactoring of the build tools could fall back to the system installs of the build tools, but that wasn't a priority for me in this PR.

If these changes make it into release, the plan is to then PR [nixpkgs](https://github.com/NixOS/nixpkgs/) with an expression for Etterna. The idea is to wrap the Etterna binary so that the game's root mount is bound to `~/.local/share/etterna` in the user's home directory by environment variables, and then to add the game's immutable installation/asset directory as an additional mount.

### DIFF
--- a/CMake/Helpers/CPackSetup.cmake
+++ b/CMake/Helpers/CPackSetup.cmake
@@ -6,7 +6,8 @@ set(CPACK_RESOURCE_FILE_LICENSE ${PROJECT_SOURCE_DIR}/CMake/CPack/license_instal
 set(CPACK_COMPONENT_ETTERNA_REQUIRED TRUE)  # Require Etterna component to be installed
 
 # Custom Variables
-set(INSTALL_DIR "Etterna")
+set(INSTALL_DIR "Etterna" CACHE STRING "Output directory for built game")
+set(ASSET_DIR "${INSTALL_DIR}" CACHE STRING "Output directory for game assets")
 
 if(UNIX)
     set(CPACK_GENERATOR TGZ)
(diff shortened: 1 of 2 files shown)


poco0317:
the changes look fine. does it work? i wont test it

fgaz:
@phantamanta44 I was looking into packaging etterna for NixOS and found this. How's your package going?

fgaz:
When that happens feel free to ping me for a review

poco0317:
although the repo activity is no indicator, it is likely that we will see a release in roughly a month. if you want to pr something to here open at any time. if you mean pr to nix or whatever just keep in mind that whatever you do for the love of god do not point a publicly available repackaging or distribution to anything other than master tags

mibmo:
> although the repo activity is no indicator, it is likely that we will see a release in roughly a month. if you want to pr something to here open at any time. if you mean pr to nix or whatever just keep in mind that whatever you do for the love of god do not point a publicly available repackaging or distribution to anything other than master tags

Why is this? wouldn't it be fine to build against the [same commit](https://github.com/etternagame/etterna/commit/8deab331e6c0e9b34eeeda363863193b26c221a4) as the newly released beta clients?

At least for the time being, till a new tag is made. :)

poco0317:
historically, people that make in-development versions of the game readily available have made giant headaches for us when the users come to us saying they installed something we didnt make and it broke

mibmo:
that's totally fair! I just miss playing the game :) I've wanted to package it for NixOS since ~last november, and found this when getting back to it again - I think I'll just create a draft PR with my derivation, so that the few that specifically search for etterna and nixos together can find it, but it won't be generally available till an actual tagged release. Would that be okay with you? I can add a disclaimer and refer to your comment.

poco0317:
if its a pr here, yes, do whatever you want. if its a pr somewhere else then im still fine with it existing unmerged as long as that isnt some annoyance to whatever maintainers exist

mibmo:
Okay, perfect! :) I've created https://github.com/NixOS/nixpkgs/pull/353956.

@phantamanta44 & @fgaz: If you two are still interested, I'd very much appreciate some feedback. Perhaps you'll share your expression?

Issue #1247, opened 2023-04-28
[Feature Request]: Add Mechanism for Deleting Songs


bluebandit21:
### Is there an existing issue for the feature?

- [X] I have searched the existing feature requests

### Describe the Feature

Add a mechanism to allow for individual songs from within packs to be deleted in-game.

### How Does The Feature Add To The Game?

It would make it possible to remove songs in-game without having to manually go in a file explorer to the relevant pack and delete them manually.

poco0317:
previously removed in 0.65.0 around these commits (primary commit lost to a squash) 22b9fd97267acb9d04d0222663825c7c9d435440 fc88c3369fbeddb25bde50d2d0048c464da5d68b

PR #1288, opened 2024-01-03
Re-Implement Song Deletion

TheROPFather:
Re-add the removed feature to delete a song permanently from the game at the song wheel using Ctrl+Backspace.

This must be enabled via the options first.

Currently this is only tested and should only work on Windows, I aim to write a function to grab the base path cross platform at a later date.

This should close https://github.com/etternagame/etterna/issues/1247


poco0317 (approved):
this will be merged along with my own modifications upon the next push to develop

Issue #1285, opened 2023-11-13
til death title color config isnt correct


poco0317:
the saturation functions are breaking the color config continuity on the screen. just try setting everything to white and witness orange

poco0317:
#1289

PR #1289, opened 2024-01-15
Saturation fix

MarcusRanger:
When you set the saturation of white (or any color with maximum value and minimum saturation), the color should still remain white because increasing the saturation should not have any effect unless the hue is defined and c.Value is less than 1.
Also cleaned up ColorToHSV as it was possible for s = -nan (division by zero)

This addresses ticket #1285


poco0317:
didnt test but this will be merged into develop on the next push

PR #1296, opened 2024-03-08
Set any 0 bpm segments to the previous segment's bpm

### LINKED ISSUE #946: Charts with certain odd BPMs listed induce a crash on load
Example: `Abe Nana(CV  Miyake Marie) - Marchen Debut! (Moskas)` from `Moskas Originals feat. PMC`

```
#BPMS:0.000=170.000
,392.000=0.000
,395.000=170.000
;
```
This is one of the simplest possible examples. What caused this crash to suddenly pop up when files like this loaded fine previously is a subtle rewrite of the function which generates non empty row data or the EndTimeAtAllRows vector.
Relevant code at the time of writing:
https://github.com/etternagame/etterna/blob/b9058c793c8004be0cac0acd667cf216fcba2c1b/src/Etterna/Models/Misc/TimingData.cpp#L1377-L1456

Until this is fixed, any file which is found to have this issue and will likely crash outputs a line in the log before it crashes (this has worked consistently for several examples so shouldn't just be missing from any of these crash scenarios).
`Found ... bps in file ... - Very likely to crash.`

niet-dev:
When I loaded a chart with bpms 170 -> 0 -> 170 in 0.70, the chart was treated as having a constant 170 bpm. To avoid timestamps going to infinity, we can treat 0 bpm segments as having the same bpm as the previous segment. Charts with `bpms.size > 1` will always have a previous segment to draw from.

Closes #946

### DIFF
--- a/src/Etterna/Models/NoteLoaders/NotesLoaderSM.cpp
+++ b/src/Etterna/Models/NoteLoaders/NotesLoaderSM.cpp
@@ -117,7 +117,7 @@ void
 SMSetBPMs(SMSongTagInfo& info)
 {
 	info.BPMChanges.clear();
-	info.loader->ParseBPMs(info.BPMChanges, (*info.params)[1], info.song->m_sSongFileName);
+	info.loader->ParseBPMs(info.BPMChanges, (*info.params)[1]);
 }
 void
 SMSetStops(SMSongTagInfo& info)
@@ -432,7 +432,6 @@ SMLoader::ProcessInstrumentTracks(Song& out, const std::string& sParam)
 void
 SMLoader::ParseBPMs(std::vector<pair<float, float>>& out,
 					const std::string& line,
-					const std::string& sPath,
 					const int rowsPerBeat)
 {
 	std::vector<std::string> arrayBPMChangeExpressions;
@@ -453,8 +452,8 @@ SMLoader::ParseBPMs(std::vector<pair<float, float>>& out,
 		const auto fBeat = RowToBeat(arrayBPMChangeValues[0], rowsPerBeat);
 		const auto fNewBPM = StringToFloat(arrayBPMChangeValues[1]);
 		if (fNewBPM == 0) {
-			Locator::getLogger()->error("Song file \"{}\" has a zero BPM. Ignored",
-										sPath);
+			Locator::getLogger()->error("Song file {} has a zero BPM. Ignored",
+										GetSongTitle());
 			continue;
 		}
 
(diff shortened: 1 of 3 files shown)


poco0317:
editing this function can have a lot of consequences. in this case, it can change calc output. it probably doesnt break much though. need to see whether or not it is possible for 2 consecutive segments to have 0 bpm, which would make this fail

niet-dev:
Sorry for my ignorance, but what would trigger the creation of a new segment when the bpm remains the same? I tested adding fakes and a stop in the middle of a 0 bpm section of a test file, but neither seemed to make a new segment.

Alternatively, if there's another direction you'd like me to go that won't affect the calc output, I'd be happy to try my hand at that instead.

poco0317:
in SMLoader::ProcessBPMsAndStops the bpms are read from the bpm string in the .sm/.ssc and without looking at that code directly i was not sure if it would interpret 2 consecutive sections of equivalent bpm as 2 segments. it isnt really possible at least in AV to do this. but the way to do it and also the way this 0 bpm thing came up was by editing the .sm in notepad and disregarding that

niet-dev:
In `SMLoader::ParseBPM`:
```
if (fNewBPM == 0) {
//			LOG->UserLog(
//			  "Song file", this->GetSongTitle(), "has a zero BPM; ignored.");
//			continue;
```
Uncommented the `continue`.

src/Etterna/Models/NoteLoaders/NotesLoaderSM.cpp
```diff
@@ -454,7 +454,7 @@ SMLoader::ParseBPMs(std::vector<pair<float, float>>& out,
 		if (fNewBPM == 0) {
 //			LOG->UserLog(
```
poco0317, line 455: `Locator::getLogger()->error("Song file {} has a zero BPM. Ignored", GetSongTitle());`

poco0317 (changes requested):

poco0317:
this is more in line with our philosophy. blame the author of the chart

poco0317:
now i dont think that outputs the actual path to the song but it would be nice if it did

poco0317 (approved):

niet-dev:
`ParseBPMs` does not seem to have access to the path directly, as `SMLoader` only stores the file name and extension. I can pass the path in via a parameter, but I'm not sure if this is a bad move.

bluebandit21:
ParseBPMs only has one caller -- that seems fine to me?

Alternatively, you could have parseBPMs throw a C++ exception and catch the exception higher up at a level where something would have access to that info.
That might actually lead to a cleaner design overall if this kind of "detect a problem and log an error" pattern was used in more places than just this :)

niet-dev:
> Alternatively, you could have parseBPMs throw a C++ exception and catch the exception higher up at a level where something would have access to that info.

This initially sounded like a good idea to me; I always feel a bit weird about changing the signature of a function unless I really know what I'm doing. However, I don't think throwing an exception would be a good choice in this case because we want the function to keep parsing the rest of the file when it finds a 0 bpm entry, rather than stopping at that point. 

I think I can do something a bit cleaner, though, by just adding the path as an attribute of `SMLoader` instead. `ParseBPMsAndStops` has a precondition of no BPM change or stop having a 0 value, and `ParseStops` has a very similar line for logging, so this route would save having to change at least 2 function signatures.

PR #1299, opened 2024-03-11
Fix Crashpad Building for CI Linux Target

bluebandit21:
For some reason, a dependency of crashpad seems to be just ...
 missing a <stdint.h> include in an obvious way.
I don't know how it ever worked before, but this fixes it now :)


bluebandit21:
Seriously no clue how it ever worked in the first place, this seems like an obvious problem to me.

:mystery:

Issue #1292, opened 2024-02-22
[META] Fix Problems Relating to macOS Notarization/Signing


bluebandit21:
# Introduction
Etterna is currently distributed as an unsigned app on all platforms, including macOS.
Windows throws some mildly-scary looking warnings up to users (which most people are conditioned to bypassing anyway), but macOS takes _many_ active measures to severely limit the functionality of unsigned apps for security reasons.

At this point, many bugs exist and have been encountered by users stemming indirectly from this problem:

## App Translocation:
By default, macOS prevents apps from accessing resources relative to the app's location.
This is to prevent a class of attacks known as "dylib hijacking", wherein an attacker maliciously replaces external resources used by an app, which can potentially cause even signed intact apps to exhibit unintended malicious behavior. [\[1\]](https://www.virusbulletin.com/virusbulletin/2015/03/dylib-hijacking-os-x)

To fix this, recent versions of macOS do not always launch apps with a current working directory equal to where the app actually lives on disk. Instead, apps are sometimes launched from a read-only dynamically created volume under /private/var/ containing the app and any internal resources. This process is known as "App Translocation", and occurs when certain conditions are met, including the launching of an unsigned app with the com.apple.quarantine xattr set (which occurs automatically when apps are downloaded through Safari) [\[2\]](https://eclecticlight.co/2023/05/09/what-causes-app-translocation/) [\[3\]](https://forums.developer.apple.com/forums/thread/724969) This breaks Etterna because we intentionally store the `Songs` and `Saves` directories relative to Etterna.app itself, instead of within `~/Library/Application\ Support` like Apple intends us to.

We currently work around this in an incredibly hacky way, by explicitly checking if we are undergoing app translocation, locating the real location of Etterna.app if so, and directly removing the com.apple.quarantine xattr from our own app ourselves before then killing and relaunching Etterna. See: 
> https://github.com/etternagame/etterna/blob/9a0b9fb94ea0c1f4d24b707bd00fad24d94d22ec/src/Core/Platform/PlatformMac.mm#L24-L69 

This workaround is bad for a number of reasons:
1) It is fragile, in that it relies upon the internal function `SecTranslocateCreateOriginalPathForURL` in the `Security` framework [\[4\]](https://opensource.apple.com/source/Security/Security-59306.11.20/OSX/libsecurity_translocate/lib/SecTranslocate.h.auto.html)
This is certainly something Apple doesn't want us doing and relies upon an internal function that could change or disappear in the future.
[...]

PR #1300, opened 2024-03-11
Automatically build Apple ARM builds in CI

bluebandit21:
Depends-on: #1298 , as without it Etterna can't be built for M1 macs.

TODO: The applications this CI generates are not signed or notarized, which will cause problems. (See: #1292 )
The next step is to automatically sign the applications made by this CI when we make a release.


nico-abram:
At a skim LGTM, great work Blue \:)

Issue #1291, opened 2024-02-13
[Bug]: ALSA Can Fail and Generate 13 GB Large Log Files


bluebandit21:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version Info

Latest available release

### What operating system are you seeing the problem on?

Linux (any distro)

### Bug Behavior

Reporting for the user `tokyostarshashtag0748` on the main Discord server.
(Arch, using pipewire for ALSA)

They encountered sound breaking for them while running the game, and later noticed that the latest log for Etterna was ludicrously large (13 GB!), filled with ```[2024-02-12 20:20:33][WARN]: ALSA: snd_pcm_wait: Broken pipe```

### Expected Behavior

Ideally, we'd be able to recover from a broken pipe situation and end up with Etterna having working sound again :)
Failing that, it's probably better to intentionally crash the game after a few repeats of the same error, since having Etterna running without sound isn't helpful, it's unlikely to recover on its own, and a log file of the same error message each frame is not helpful either.

### Reproduction Steps

Unknown

### Anything else?

The actual error message and consequent logging is happening within
https://github.com/etternagame/etterna/blob/9a0b9fb94ea0c1f4d24b707bd00fad24d94d22ec/src/arch/Sound/ALSA9Helpers.cpp#L359-L369

The ALSA documentation for the snd_pcm_wait command is available at https://www.alsa-project.org/alsa-doc/alsa-lib/group___p_c_m.html#gad4d53d58b996a7cd9a5cbf1710b90375

bluebandit21:
The "Intentionally crash the game" approach is pretty simple -- see if we get the same error code returned for, idk, 10 sequential calls of the function; crash if so with an appropriate message, resetting that counter if it ever succeeds.

I might try to look into ALSA stuff sometime soon to see if it's relatively simple to fix the broken pipe after it occurs.

Unfortunately I don't have a system on which to test changes/reproduce the bug.

bluebandit21:
See: https://discord.com/channels/339597420239519755/339597420239519755/1206772757287804929, https://discord.com/channels/339597420239519755/339597420239519755/1206765186103382047 for context.

Vezxe:
Same issue

StarbotArc:
I have also experienced it, but I feel like the game shouldn't just crash because of this issue because that will happen while a player is playing the game and will cause more issues. I think really it should tell the user upfront when they aren't playing a file that the audio does not work.

[...]

PR #1304, opened 2024-04-15
Consolidate consecutive duplicate logger lines

StarbotArc:
Basically, no more big file sizes from repetitive logs. This pull request exists for issue #1291.

### DIFF
--- a/src/Core/Misc/PlogLogger.cpp
+++ b/src/Core/Misc/PlogLogger.cpp
@@ -87,7 +87,7 @@ PlogLogger::PlogLogger() {
 
 void PlogLogger::log(Core::ILogger::Severity logLevel, const std::string_view message) {
     const std::lock_guard<std::mutex> lock(this->mutex);
-    if (this->last_msg.empty()) {
+    if (!this->last_msg.empty()) {
         if (message == this->last_msg) {
             this->count++;
         } else {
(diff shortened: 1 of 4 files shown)


nico-abram:
Why is last_msg a pointer? Can't we use a std::string directly and use an empty string instead of null?

StarbotArc:
Sorry, I don't have crazy on-hand experience with C++, I actually write more C code, so I resorted more to pointers instead. If you think that's a better idea, I would say go for it.

nico-abram:
LGTM

Only nits like I mentioned on discord are 

(1) maybe we can unlock the mutex before the call to plog by making a string copy
(2) I think we can remove the `if (this->last_msg.size() > 0)` condition, the initial empty string should work fine on the `if (message == this->last_msg)` check

src/Core/Misc/PlogLogger.cpp
```diff
@@ -84,7 +86,19 @@ PlogLogger::PlogLogger() {
 }
 
 void PlogLogger::log(Core::ILogger::Severity logLevel, const std::string_view message) {
-    PLOG(PlogLogger::convertSeverity(logLevel)) << message;
+    const std::lock_guard<std::mutex> lock(this->mutex);
+    if (this->last_msg.empty()) {
```
poco0317, line 90: if last_msg is empty, then check to see if the new message is equal to that empty message? else if it isnt empty, set the last message to the new message? youve reversed the logic with the most recent commit

poco0317 (changes requested):

src/Core/Misc/PlogLogger.cpp
StarbotArc: Yeah, that's a major mistake, I'll get to fixing that.

poco0317 (approved):

poco0317:
superceded by #1305

PR #1305, opened 2024-04-20
Fix duplicate log suppression pr :)

bluebandit21:
Should fix problems of #1304 

Note -- in order to prevent acquiring a global mutex between multiple threads, I exploit the fact that a spammed log message would happen on the same thread, letting us use thread_local variables instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@
 .vscode
 .vs
 .idea
+.kdev4
+*.kdev4
 /[Bb]uild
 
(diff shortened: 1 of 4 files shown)


poco0317:
replaced by 82a5e44653545910536f6ec318340191bbbdff51 because I think the amount of effort going into consolidating log lines has gotten lost in the sauce. it's better for now to restrict log file sizes and be reasonable about the amount of logging we do. if a log file ends up hitting a 64mb cap, then theres obviously a problem and we can look at how that happened in the first place

PR #1307, opened 2024-05-01
Add PB and Diff display in Til' Death Playlists and add Refresh Button if you don't have songs

yeslyko:
Also made the song title more wide so it wouldn't look ugly for medieval LG Flatron users


poco0317:
d03e95ab2b66e97f91378e6003e6c375eff8cdd1 merged

PR #1308, opened 2024-06-03
Build docs: openssl-static -> openssl-devel on Fedora

kangalio:
`openssl-static` doesn't exist anymore. The new name is `openssl-devel` now https://packages.fedoraproject.org/pkgs/openssl/openssl-devel/

### DIFF
--- a/Docs/Building.md
+++ b/Docs/Building.md
@@ -60,7 +60,7 @@ While most dependencies for macOS and Windows are included in the repo, there ar
 For Pop_OS!, install the `Debian` dependencies, then refer to the [Pop_OS! Notes](#pop_os-notes) section below.
 
 - Debian: `apt install build-essential libssl-dev libx11-dev libxrandr-dev libcurl4-openssl-dev libglu1-mesa-dev libpulse-dev libogg-dev libasound-dev libjack-dev`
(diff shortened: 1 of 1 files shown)


poco0317:
? https://packages.fedoraproject.org/pkgs/openssl/openssl-static/index.html

bluebandit21:
@poco0317  --

Screen Shot 2024-06-03 at 11 41 23 PM


Screen Shot 2024-06-03 at 11 41 35 PM

poco0317:
you've send screenshots of the same pages i read. that's not particularly helpful. im interested in whether its useful information to add different dependency names depending on os. linux continues to prove itself as insane and annoying to deal with

bluebandit21:
Pointing out that openssl-static isn't being used anymore since Fedora 36 :)
(You seem confused by Poco's PR, so thought I'd add some context)

kangalio:
> you've send screenshots of the same pages i read. that's not particularly helpful. im interested in whether its useful information to add different dependency names depending on os. linux continues to prove itself as insane and annoying to deal with

Fair point. But Fedora 35 has been EOL since December 2022 and other Fedora openssl guides on the internet don't mention the old package name either, so it should be fine

PR #1316, opened 2024-08-03
Unsuck unpitch rates

graemephi:
This turned out to be easier than I expected

The main thing is not letting the left and right channels get time stretched/squeezed independently, which you can do with [this nonsense](https://en.wikipedia.org/wiki/Joint_encoding#M/S_stereo_coding). Less obtusely you can think of it as doing the algorithm on the average of the left/right channels instead of each independently, and then applying the change it wanted to make to both channels. And actually you could do exactly that, but I don't, because it helps masks an annoying issue to do with the way uncorrelated audio mixes, which makes certain noisey songs sound fluttery

The algorithm this thing uses still has a more fundamental issue which is on uprates it will just straight up skip beats. There's really no good reason for this but I don't think it can be fixed without just rewriting it entirely


poco0317:
this certainly sounds different than before

Issue #759, opened 2020-05-13
Considering hitsounds


kangalioo:
Etterna currently doesn't have hitsounds like in osu. However, there is frequent demand for exactly that. This demand mostly comes from people who are used to osu.

The Etterna community tends to look down on osu and its community. This is probably why hitsounds are everything but a priority in this game (read: people actively _don't want_ to even have the _option_ to enable hitsounds).

However, I do think it should at least be considered to be added **as an optional feature**. The only objective reason _against_ adding optional hitsounds to Etterna is the work involved.

The upside is better accessibility for osu migrants, which would lead to more players. Plus, it might make give a better look on the community as a whole.

MinaciousGrace:
my stance on stuff like this (as you know) is generally: will merge if i dont have to do anything including spending time testing it and someone can confirm it adds no overhead to gameplay if disabled

though i will probably chafe if it looks like its too much of a hack job

poco0317:
on another note people like to complain that hitsounds added through noteskins or other methods are delayed and useless anyways

PR #1322, opened 2024-09-24
Add and implement hitsound option

CreoSm:
Add feature https://github.com/etternagame/etterna/issues/759 which will be **disabled by default**.

This option is located under the Graphics/Sound options and when enabled will play the sound file (if available) "Player hitsound.ogg" when hitting a note successfully (no cb). The sound must be located in the Sounds folder of whatever theme is being used, or fallback.

I have not provided a sound in this pr to be the default etterna hitsound, if this is implemented someone would have to make one, or players will just need to add one themselves.


StarbotArc:
Works out pretty well, if it could be somehow moved into noteskins and the judgement (or lack of one) that allows it to play could be adjust it, I think it'd work out perfectly.

CreoSm:
> Works out pretty well, if it could be somehow moved into noteskins and the judgement (or lack of one) that allows it to play could be adjusted, I think it'd work out perfectly.

I might rewrite this to make it work that way. I WAS going to add CB sounds but had some issues with it

PR #1324, opened 2024-09-28
Default WITH_CRASHPAD to OFF instead of ON in CMake

bluebandit21:
Crashpad is only really useful in the context where multiple
 people are using the same build and remote debugging is
 desirable, i.e. distribution builds.

For people building locally on their own machine, Crashpad is
 extremely unlikely to be helpful, and creating the environment
 where crashpad can actually be used involves the horror of
 Google's depot_tools and such...

It seems more reasonable to make users explicitly opt *into*
 crashpad, rather than having to opt out, especially when the
 error message produced when trying to build with the past
 default settings was more likely to get users to try to
 install Crashpad to fix the problem rather than simply
 disabling it.


bluebandit21:
Note: I attempted to modify all of the existing documentation mentioning Crashpad to reflect this change, but it's possible I missed something somewhere.

bluebandit21:
Hmm, maybe don't merge this yet, regarding the discussion in the dev Discord :)

PR #1344, opened 2025-01-07
Replace unweildy COMPILE_ASSERT macro with static_assert

bluebandit21:
When this codebase was originally written, there wasn't a built-in
 cross-compiler way to add compile-time asserts, so the original
 authors did something clever by forcing the compile-time
 instantiation of a undefined template class if a precondition was
 not met.

However, this produces a wall of insane template errors all wrapped
 within macro expansions that is completely unintelligible if you
 haven't seen it before.

Since C++11 we have static_assert for the same purpose, and can
 produce nice diagnostic strings as an added bonus :)


bluebandit21:
Old:
Screen Shot 2025-01-07 at 11 28 08 AM


New:
Screen Shot 2025-01-07 at 11 21 02 AM

PR #1351, opened 2025-04-01
Pumpual Handswitching Deditechcion Calculation

StarbotArc:
Essentially detecting if your hands should have been switching or something, then making sure it doesn't violate arbitrary rules.

Feel free to rag on it to rag on it, and please give some form of feedback.


poco0317:
to start, there is a random print statement added and there are unclear changes to enumhelper

StarbotArc:
image
> ctrl + f
> print
> "i am pump"
> start dying of laughter

Anyways, EnumHelper was preventing g++ from compiling at some point, so it's just there. I'll both commit it out in a second.

StarbotArc:
Okay, now I actually remember, it was that funny thing you told me about errors coming from having a string equivalent to an enum, and before I actually asked I commented out the undefined template thing for GCC. Anyways any moment now.

StarbotArc:
oh shoot, that's to the master branch

StarbotArc:
Modifications from pbsm:

Addition of `UlbuFiveKey`, for 5k styles such as pump or pop'n. (UlbuFiveKey.h & MinaCalc.cpp)
Addition of dependent mod `HandSwitchMod`, that processes notes and decides to value whether this would warrant a switch in hand. (HandSwitch.h)
Addition of `HandSwitch` enum for `CalcPatternMod` along with its string equivalent in `CalcPatternModNames`. (NoteDataStructures.h & GameConstantsAndTypes.cpp)
Incremented version for the calculator. (MinaCalc.h)

(there's probably more idk)

src/Etterna/MinaCalc/UlbuFiveKey.h
```diff
@@ -0,0 +1,228 @@
...
+			case Skill_JackSpeed:
+				break;
+			case Skill_Chordjack:
+				break;
+			case Skill_Technical: {
+				*adj_diff =
+				  _calc.init_base_diff_vals.at(hand).at(TechBase).at(itv) *
+				  pmod_product_cur_interval.at(ss) * basescalers.at(ss);
+
+				*adj_diff += _calc.init_base_diff_vals.at(hand).at(MSBase).at(itv) *
```
poco0317, line 107: the way this adj diff is being set up is weird because you are double applying a pmod by including it both in the pmod map and here. also msbase was never meant to be used and is more of a chordjack thing. but if it produces numbers i guess it works

src/Etterna/MinaCalc/Dependent/HD_PatternMods/HandSwitch.h
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	void advance_sequencing(const float& ms_now, const unsigned& notes)
+	{
+		_eml--;
+
+		// prevents patterning like this, as it's a different field of tech.
+		// 11011
+		// 00100
+		if ((notes & 0b11) >= 0b11 && (notes & 0b00011) >= 0b00011)
```
poco0317: this line is flagged as a redundant code warning. youve really written the same statement twice here. `0b00011` does not allow you to specify the bit width of a binary number

src/Etterna/MinaCalc/Dependent/HD_PatternMods/HandSwitch.h
```diff
@@ -0,0 +1,163 @@
...
+		if (notes == _last && notes != 0b00100)
+		{
+			no();
+			_eml = 0;
+			return;
+		}
+		if (notes & 0b00100)
+		{
+			// chord that has both hands and middle lane involved should not be counted as handswitch.
+			if ((notes & 0b11) && (notes & 0b00011))
```
poco0317: this is flagged with a redundant code warning because `0b00011` is the same thing as `0b11`

src/Etterna/MinaCalc/Dependent/HD_PatternMods/HandSwitch.h
StarbotArc: Yeah, that should really be ``0b11000`` but my brain thinks my high bits (4 + 5) on the rights, I'll go fix that up tomorrow.

src/Etterna/MinaCalc/UlbuFiveKey.h
StarbotArc: I mean I can just remove the double instance and see if it works correctly with ``MSBase``, but this is a moment where I get an info disconnect just because I am not too familar with Ulbu's pmodding jazz. I'll check up on it tomorrow as well.

src/Etterna/MinaCalc/Dependent/HD_Sequencers/CMakeLists.txt
```diff
@@ -4,4 +4,4 @@ list(APPEND HD_SEQUENCERS
     "OHJSequencing.h"
     "RMSequencing.h")
 
-target_sources(Etterna PUBLIC ${HD_SEQUENCERS})
\ No newline at end of file
+target_sources(Etterna PUBLIC ${HD_SEQUENCERS})
```
poco0317, line 7: why

Themes/Rebirth/BGAnimations/ScreenSelectMusic decorations/calcDebug.lua
```diff
@@ -962,6 +965,7 @@ local modnames = {
     "gstrea",
     "gchstr",
     "gbrack",
+    "hsw",
```
poco0317, line 968: without the changes in this file existing in the Til Death equivalent (in the exact same format so its a simple copy paste) the Til Death calc debug doesnt work correctly when scrolling through the pmod graphs

src/Etterna/MinaCalc/Dependent/HD_Sequencers/CMakeLists.txt
StarbotArc: yeah i don't understand why this is a commit either

poco0317 (changes requested):

Themes/Rebirth/BGAnimations/ScreenSelectMusic decorations/calcDebug.lua
StarbotArc: Alright, I can just go do a small commit for this one real quick.

StarbotArc: Okay, Til Death now has it too.

src/Etterna/MinaCalc/UlbuFiveKey.h
poco0317: the mod enums defined in the pmod map are automatically applied per interval to the adj_diff.

technical in all the different ulbu implementations is special in that it assigns a value to it, replacing that adj_diff and ignoring the nps based adj_diff. but note that it includes `pmod_product_cur_interval` which is the value of the pmod multipliers combined for that interval. you do the same here, which is fine - but you are also applying HandSwitch a second time by adding to the adj_diff with some funny fudgy looking math also done to it

StarbotArc: Oh, okay, so you're essentially telling me here that I could just simplify that and get virtually the same result?

poco0317: yes, you have several options
 - remove HandSwitch from usage in this function
 - remove HandSwitch from the pmods map
 - dont remove it and make absolutely sure that you want it to be kind of doubly applied and it makes sense to do that
 - make Technical replace the adj_diff value like it is now but dont refer to the pmod_product_cur_interval, which will mean that all pmods are disabled in tech for 5k except for this specific hardcoded formula

if you are using calc debug while messing with this, refer to the TotalPatternMod graphs on the top part of the screen. they are basically the pmod_product_cur_interval (for tech, it is based on TechBase while others are based on NPSBase)

StarbotArc:
I lied, I got bored and decided to just get fixing anyways.

poco0317 (approved):

PR #1369, opened 2025-06-04
Initialize depot_tools before using it to avoid CI error, upgrade GHA Windows runner for same reason

bluebandit21:
Before this change, the CI was failing with the message
```
python3_bin_reldir.txt not found. need to initialize depot_tools by
running gclient, update_depot_tools or ensure_bootstrap.
```
in the crashpad_init gn_configure phase.

This commit simply runs `update_depot_tools` before running gn,
also including the necessary CMAKE change to locate the binary

It also upgrades the Windows GHA runner from Windows-2019 to Windows-2022, as without this change update_depot_tools fails in a strange way in the midst of running a Visual Studio 2019 command.


nico-abram:
My cmake is not great but LGTM

nico-abram (approved):

PR #1370, opened 2025-06-09
SetPage is now a function for Lua.

StarbotArc:
I heard you wanted to set pages...

Changed visibility of private method `setPage` in `DownloadablePackPagination` to public, as to allow the Luna implementation see and use it. (DownloadManager.h)

Added function `SetPage` to Lua, as to allow more advanced page navigation with it. (DownloadManager.cpp)

### DIFF
--- a/src/Etterna/Singletons/DownloadManager.h
+++ b/src/Etterna/Singletons/DownloadManager.h
@@ -224,6 +224,12 @@ class DownloadablePackPagination
 	// get all the packs that are cached
 	std::vector<DownloadablePack*> getCache() { return results; }
 
+	// had to move this to public to let SetPage see it.
+
+	// move to n page and send packs to lua function
+	// possibly invoking a search request
+	void setPage(int page, LuaReference& whenDone = EMPTY_REFERENCE);
+
 	// move to next page and send packs to lua function
 	// possibly invoking a search request
 	void nextPage(LuaReference& whenDone = EMPTY_REFERENCE)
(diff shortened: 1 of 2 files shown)


src/Etterna/Singletons/DownloadManager.h
```diff
@@ -224,6 +224,12 @@ class DownloadablePackPagination
 	// get all the packs that are cached
 	std::vector<DownloadablePack*> getCache() { return results; }
 
+	// had to move this to public to let SetPage see it.
```
poco0317: this is a comment that belongs in the description of a pr because nobody is going to know what it means

poco0317 (changes requested):

src/Etterna/Singletons/DownloadManager.h
StarbotArc: Alright, I'll edit my PR description to mention changes.

StarbotArc: Okay, the description should have more specifics on what modifications I have made in the codebase, and also put emphasis that I swapped visibility to `setPage`.

poco0317: this comment should not be here

poco0317 (changes requested):

src/Etterna/Singletons/DownloadManager.h
StarbotArc: Comment now removed.

poco0317 (approved):

Issue #1041, opened 2021-08-28
"SmoothLines" option breaks D3D fullscreen screenshots


Jolomann:
**Describe the bug**
ok can i just preface this by saying this is really stupid and i hate that a long standing, etterna specific bug (at least since .57 or .60 i believe?) is caused by this

so, taking a FULL SIZE, FULLSCREEN screenshot in D3D has been broken to some degree for a while. it has only worked while holding shift for some users. today i spoke to caliber who never had this issue *in d3d fullscreen* so i started investigating

i tried using his `Preferences.ini` file and then it literally just worked. comparing our files i narrowed it down to
`SmoothLines=0` causing it to work

**To Reproduce**
boot the game up with `SmoothLines=0` and `VideoRenderers=d3d` in `Preferences.ini`, make sure you're in fullscreen, hit printscreen, and it takes a screenshot

do the same but with `SmoothLines=1`, and now it doesn't work, though the game still hangs for about 200ms, albeit without any screenshot sound

if the issue isn't present, see below

**Additional Context**
confusing part here; caliber didn't have the issue regardless of `SmoothLines` being on or off. since i have the issue with an NVIDIA GPU and caliber doesn't with an AMD GPU, i checked with other people who have different models and it doesn't seem to be strictly an NVIDIA issue

results are:
NVIDIA GeForce GTX970: Issue present
NVIDIA GeForce RTX2070 SUPER: Issue present
AMD Radeon HD7950: Issue pressent
AMD Radeon RX480: Issue not present

so it doesn't appear to be a brand specific driver issue, but might hint at newer AMD models maybe being immune to this? not sure if this can really be conclusive, may need more AMD people to test

**Version**
Etterna 0.70.3 (but really this should be present on all versions since roughly .57 or .60)

**OS**
Windows 10

Jolomann:
tl;dr if you play fullscreen with D3D and want to take big screenshots by pressing printscreen, turn off smoothlines

(not applicable to all graphics cards though)

poco0317:
thats cool but does pressing / or left+right do anything instead

Jolomann:
sorry should've specified, / and left-right works as an alternative but both of them are exclusive to evalscreen

poco0317:
the issue is narrowed down to the input of MultiSampleType here https://github.com/etternagame/etterna/blob/08773a50b4d42f4b787304e0d74483ce7520327d/src/RageUtil/Graphics/RageDisplay_D3D.cpp#L537-L538

[...]

Issue #661, opened 2019-12-09
Models fail to color on D3D


poco0317:
for example, a noteskin that has a texture and model set up will be fully white if using D3D.
If using OpenGL, the model works a bit more as you would expect.

I may not be using the right terms here, but you get the idea. (it's a Def.Model in Lua)

Checked on Etterna 0.57, .60, .61, .62, .63, .64, .65, .66, .67, Stepmania 5.0.12, 5.0.9 and this issue is still present.

Windows specific since it's a D3D thing.

PR #1371, opened 2025-06-12
Shader support in RageDisplay and Actor

convextriangle:
**Not ready for merging (yet); created so stuff isn't lost**

This PR adds support for adding vertex and fragment shaders (for D3D9 and OpenGL separately) to any Actor.
At the moment it's only a prototype/proof-of-concept mainly for D3D, to do:

General
--
- [ ] Remove/replace old prototype code (18880bde9b84f8eb5b213ce7611b868bab6d87f9 and everything before it)
- [x] Add some sort of `std::chrono::high_resolution_clock` for shader time parameters...
- [ ] Add Lua API to initialize and set shaders to Actors (and possibly some extras like setting uniforms or constants)
    - also add getting current renderer and shader profiles in Lua
- [ ] Add loading uniforms (so far, only hardcoded WVP and time passing exists in D3D)
    - the current "push uniforms together with the shader" is comically inefficient
- [x] Graphics settings changes -- doublecheck if things are broken for OpenGL/D3D (e.g., when switching fullscreen<->windowed, the thingy blacks out)
    - should be resolved in 27fdb4b76dcbfbcca4f7af1ca4732a99ec69e8d5?
- [ ] 3D models, such as noteskins like [Metal](https://storage.etternaonline.com/resources/noteskins/uswmetalett-O9sA.zip)
    - related to #661 (test case for future changes)
- [ ] Investigate possible pooling/arena things at least for default shaders <br>(RageShaderWeakRef is likely too much indirection, replace with enums or something)
- [x] Cleanup shader cleanup (track currently set shaders in a sane way)
- [ ] Investigate performance / compatibility with different operating systems / others
- [ ] Other things that either someone else will bring up or I'll remember later
- [ ] Game crashes after failing a chart (`Exception thrown: read access violation. this-> was 0xFFFFFFFFFFFFFFD7.`)
- [ ] changing tabs on fullscreen breaks things

D3D
--
- [ ] #1041 still exists
- [ ] Move away from FVF for D3D (it's possible that for most actors one can just port the default GLSL shaders, likely would reduce a lot of switching between shaders and FVF)

OpenGL
--
- [ ] remake RageDisplay_Legacy into an actual class (as for RageDisplay_D3D)


convextriangle:
the prototype introduces far too many changes that would require breaking a lot of current handling - [closing for separation to an isolated display](https://github.com/convextriangle/etterna/tree/unstable_d3d)

PR #1374, opened 2025-06-16
Tell Cmake we want to use OpenSSL 3.5 specifically

bluebandit21:
Without this change, CMake was locating a OpenSSL 3.4 package on my computer and consequently failing later on.


bluebandit21:
Don't merge this yet until tests pass; building is currently hecked because the CI doesn't have OpenSSL 3.5, working on fixing that right now :)

PR #1381, opened 2025-07-13
Add new filter "4K Only"

NebuPookins:
This hides all 8K (aka Doubles) chart from the song selection wheel. This is convenient for people who do not have the hardware to play double.

Tested in both "Rebirth" and "Til Death" themes on Linux.


poco0317:
dance is not the only game mode which exists, so this doesnt make sense for something such as solo - which is 6 key and doesnt have a doubles mode. it might be better to just go directly to keymode filters whenever we integrate all keymodes for all gamemodes onto the same wheel

NebuPookins:
Thanks, I hadn't thought of the 6key issue. I find it inconvenient to have to filter out the 8k songs, so I'd like to see something equivalent to this feature come out without waiting for integrating all keymodes onto the same wheel (which I assume won't arrive for a long time?), so I'm open to suggestions on how to make this work. I can try to detect the game mode and only show it when it looks like you're playing DDR?

StarbotArc:
We already had a way, but the themes we include (excluding bare-frames) don't force you to select a certain play style for it and instead just shows every type of chart for a particular game-style.

NebuPookins:
I found another way to implement this; see https://github.com/etternagame/etterna/pull/1382

PR #1383, opened 2025-07-29
Beta DirectX 12 display (for shader usage in Actors)

convextriangle:
The previous prototype #1371 had too many breaking changes, isolating to this draft.

General
--
- [x] Remove/replace old prototype code (18880bde9b84f8eb5b213ce7611b868bab6d87f9 and everything before it)
- [ ] Add some sort of `std::chrono::high_resolution_clock` for shader time parameters...
- [ ] Add Lua API to initialize and set shaders to Actors (and possibly some extras like setting uniforms or constants)
    - also add getting current renderer and shader profiles in Lua
- [ ] Add loading uniforms (so far, only hardcoded WVP and time passing exists in D3D)
    - the current "push uniforms together with the shader" is comically inefficient
- [ ] 3D models, such as noteskins like [Metal](https://storage.etternaonline.com/resources/noteskins/uswmetalett-O9sA.zip)
    - related to #661 (test case for future changes)
- [ ] Investigate possible pooling/arena things at least for default shaders <br>(RageShaderWeakRef is likely too much indirection, replace with enums or something)
- [x] Cleanup shader cleanup (track currently set shaders in a sane way)
- [ ] Investigate performance / compatibility with different operating systems / others
- [ ] Other things that either someone else will bring up or I'll remember later
- [ ] changing tabs on fullscreen breaks things

D3D
--
- [ ] #1041 still exists


convextriangle:
for posterity - postponing until october, not sure if i'll have enough time
at the current prototype state, these remain (mainly):
- uploading draw command & texture data
- RageModelVertex
- testing with all actors
- profiling for FPS?
- other things (maybe mesh shaders or something compute-ish for cool effects)
- all unsupported / `assert(false)`-type functions in `Display::Display`
- synchronizing without CPU/OS-based fence event?
- Linux/MacOS support for `Display::Display`

tl;dr burnout/exhaustion :)

convextriangle:
soon™

PR #1384, opened 2025-08-02
Add more debug info when hand_counts assertion fails.

NebuPookins:
I've been getting some assertion failures like:

```
[FATAL]: nri.hand_counts[left_hand] + nri.hand_counts[right_hand] == nri.row_count
```

which I think might be related to PumpItUp charts. I don't fully understand what the hand_counts logic is doing. But in the meantime, I've updated the assertion message to print out some more info. In particular, I made it print out the name of the file that's triggering the assertion failure. This required adding an extra parameter to several methods so that the filename is passed up the call chain.

The new error message looks like this:

```
[FATAL]: Hand count mismatch detected!
  File: /Songs/StepMix 2 Winners/V1 (for SM and PiU)/V1 (for SM).sm
  Keycount: 5
  Row notes (binary): 00000000000000000000000000000100
  Row count: 1
  Left hand count: 0
  Right hand count: 0
  Left hand mask: 00000000000000000000000000000011
  Right hand mask: 00000000000000000000000000011000
  All columns without middle: 00000000000000000000000000011011
  Row time: 0.579151
  Interval: 1, Row: 0
  Expected: 0 == 1
```

The main benefit of this is that now that we have a filename, it's possible to inspect the file to see what may have gone wrong. So far, every instance I've seen of triggering this assertion failure seems to be related to having a PumpItUp Chart.

### DIFF
--- a/src/Etterna/MinaCalc/MinaCalc.cpp
+++ b/src/Etterna/MinaCalc/MinaCalc.cpp
@@ -50,7 +50,8 @@ TotalMaxPoints(const Calc& calc) -> float
 auto
 Calc::CalcMain(const std::vector<NoteInfo>& NoteInfo,
 			   const float music_rate,
-			   const float score_goal) -> std::vector<float>
+			   const float score_goal,
+			   const std::string& filename) -> std::vector<float>
 {
 
 	InitializeKeycountLogic();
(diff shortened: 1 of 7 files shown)


poco0317:
This change isn't acceptable because it's making a breaking API change. Without the additional string parameter, it would be fine. However, this pr is more of a report of an actual programming problem. So I went ahead and fixed it: https://github.com/etternagame/etterna/commit/164bd55187fd7e29c978432bd4982c39dfe1c795

NebuPookins:
Thanks for the fix.

I wasn't aware that this API is "public". Are there external third parties that call into these APIs?

poco0317:
Everything behind define `PHPCALC` is used as a standalone version of the calc for php c bindings. The MinaSDCalc signature is used for a java interface for another website implementation here https://github.com/poco0317/calc-jni as well as other unauthorized duplicate minacalc repos for rust/c++ use in other projects

PR #1396, opened 2025-11-03
Add aarch64-linux support

marcin-serwin:


poco0317:
once the workflows dont suck in general i would like for aarch64-linux to be its own workflow. it will wait to be merged until that, because i cant verify whether this just breaks everything else for any reason. that will happen before the next tag/release

PR #1406, opened 2025-11-20
Alt+Enter Window Fix

StarbotArc:
~~This should fix how the window handles context resetting, because apparently we've been doing it **after** the window got resized, forcing it to never really update on exclusive fullscreen.~~

~~My fix is literally just doing it when the keybind is pressed.~~

(this is stupid, i found out after.)


poco0317:
it isnt broken and the explanation you gave here doesnt make sense based on what you changed. find the reproducible problem and fix that

PR #1408, opened 2025-12-01
Replace SaveBMP implementation with stbi_write_bmp

atomicf32:
Compiled and tested on archlinux


poco0317:
it isnt even possible for this code to be reached without modifying the c++ but sure why not

PR #1411, opened 2025-12-10
Just homework

aaddaad:
I'm sorry for submitting this request; this is just a course assignment. We made the following changes to the project:
1. Added Simplified Chinese translation file
Added the translation file Themes/_fallback/Languages/zh-cn.ini.

2. Refactored README document structure
Added a concise project description below the title.

Added a detailed navigation directory for quick access.

Reorganized content from a linear layout into modular categories.

3. Optimized readability of GameState.cpp code
Refactored the JoinPlayers function, removing redundant temporary variables.

Simplified the GetNumSidesJoined function to return directly based on boolean members.

Ensured that no external logic was changed, only readability and maintainability were improved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,85 +1,67 @@
-<p align="center">
-    <img src="Docs/images/etterna-logo-dark.svg" width=250px>
-</p>
+# Etterna - Open Source Rhythm Game
 
-<p align=center>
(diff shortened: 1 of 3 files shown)


poco0317:
we cant accept this in its current state because
- the change to the code includes comments none of the active developers can read
- the readme rewrite is unfinished and not something we need

on its own, the zh-cn language file might be acceptable. i would want to know if you can actually load it ingame. if you can, then we can take it

PR #1428, opened 2026-04-07
Patch up macOS compilation on CI

convextriangle:
seems that some frameworks (AppKit and Foundation) were missing+


convextriangle:
ew

PR #1437, opened 2026-05-18
Handswitch pmod update and TechBase implementation for UlbuBase

StarbotArc:
`HandSwitch.h` - A updated version since `TechBase` implementation, with it being more balanced by side effect (notably: it doesn't gain a significant amount from specific patterning, but does recognize them), and resembling a 'normal' pattern mod (i hope).
`UlbuBase.h` - Pulling code from the 4-key calc. implementation to run on `UlbuBase`, allowing children to have this implementation as well.
`UlbuFiveKey.h` - I don't know what actually changed to be honest else than how tech is rated and HandBalance being added just for more things to give tech rating.

ps: do i capitalize pmod PMod or Pmod?


poco0317:
PatternMod PMod

PR #1445, opened 2026-07-06
switch from websocketpp to libcurl-ws

convextriangle:
It has a transitive dependency on old OpenSSL which breaks CI.


poco0317 (commented):
this is curl 8.21.0, not 8.2.1

convextriangle:
> this is curl 8.21.0, not 8.2.1

typo :(

PR #1454, opened 2026-07-28
Added clicking cdtitle filters by chart author

Gtexchus:
'Til Death:
* Click on a cdtitle to search by the chart author. Search tab is updated accordingly
* Click a cdtitle again (without typing in the search tab) to remove the search filter from the wheel
* Typing in the search tab after clicking a cdtitle means the next cdtitle click will NOT remove the search filter, and will instead search for the chart author

Rebirth:
* Click on a cdtitle to search by the chart author. Search tab is updated accordingly. All other search fields are reset
* Click a cdtitle again (without updating the author field) to remove the search filter from the wheel
* Updating the author field in the search tab after clicking a cdtitle means the next cdtitle click will NOT remove the search filter, and will instead search for the chart author


poco0317:
the third line of each description is confusing but i guess it is technically correct

Issue #1348, opened 2025-03-29
[Bug]: No video playback on linux


dmitsuki:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Version Info

Compiled from develop

### What operating system are you seeing the problem on?

Linux (any distro)

### Bug Behavior

No videos play on Linux. Currently on arch linux. All you get is a black screen, regardless of in game settings. Running through wine you get the expected behavior. 

### Expected Behavior

Video plays with song

### Reproduction Steps

Run the game on linux, either from the provided binaries, or compiled from source

MandL27:
Reproduced on 0.74.4 on Bazzite 42, both natively and from an Arch distrobox. 

[Logs](https://github.com/user-attachments/files/20743179/2025_06_14-21_33_20.log) don't have anything that immediately jumps out to me; in fact they were completely silent between 21:33:31 (began file) and 21:34:53 (quit).

hubret:
after doing some ai assisted digging (since i am new here), i found that this is caused by ffmpeg not being linked on linux pre-2022 due to not having a compiled library for linux. the bundled ffmpeg is v2.1.3 from 2014, for windows 32 bit. (theres also a 64 bit lib with unknown version, but unsure if it works) https://github.com/etternagame/etterna/pull/1046#issuecomment-1243130126

the lib files end with .dll.a because this ffmpeg build's target was windows 32 bit. unsure if they actually would have worked on linux.

ive tested with system ffmpeg libs, and i got bg video to work after updating the api to v4.0+ in MovieTexture_FFMpeg, since it was last using the api of the bundled 32 bit ffmpeg (unsure about if 64 bit version would work)

ffmpeg has had [autobuilds](https://github.com/BtbN/FFmpeg-Builds/releases) (listed on its official website) since march 2024, so IIUC we can use these without building ffmpeg from source for each platform.

so bundled libs just need to be updated and separated into platform specific lib sets. the CMakeLists can select libs based on the values of WIN32, APPLE, and UNIX.

and if these are updated, DirectShow implementation (and Null) is maybe not needed?

EDIT: it seems like some of the new ffmpeg .so files are too big to be pushed through without LFS.

cc @bluebandit21

bluebandit21:
Hey, I'm so sorry to take so long to respond to this, been busy with IRL stuff and haven't been very active with Etterna development for a while here.

[...]

PR #1458, opened 2026-08-01
Use precompiled ffmpeg 2.1.3 for x64 Linux

DashDashDashDashDash:
This fixes https://github.com/etternagame/etterna/issues/1348 
Opted not to try having Linux build ffmpeg since it requires an outdated GCC and outdated libraries to properly compile, and it avoids touching the existing ffmpeg API.
Adds 18.6MB of binaries to the repo. Build was sourced from https://ffmpeg.org/releases/ffmpeg-2.1.3.tar.gz and was compiled similarly to its Windows counterpart, but without gnutls, libbluray and libass support. As far as I'm aware, the game doesn't use any of the features exposed by these components.
Tested on Debian 13 in a VM and on Arch Linux in real hardware.

Screenshot From 2026-08-01 13-47-22


PoshestPoro:
Also tested on Gentoo on my end.

poco0317:
the build must pass

DashDashDashDashDash:
Wasn't able to find any cross compilation toolchains for that old of a Debian version. Opted to have ARM Linux completely ignore the precompiled binaries since it's an optional dependency and I doubt there are lots of clients running ARM in the first place

PR #1465, opened 2026-08-13
Replace std::vector::reserve with fixed buffer

ahigerd:
This fixes a crash on startup when bounds checking is enabled (i.e., debug builds) when trying to dereference `&buffer[0]`.

Switching to resize() would have worked too (but the clear() call later would still need updated), but std::array exists for this kind of use case, so might as well use it.


poco0317:
so actually the reason for this was that msvc said the stack is unusually big for the thread when using std array

ahigerd:
Oh. Well, then, I'll switch it back to vector and just not use reserve().

PR #1418, opened 2026-03-15
RageDisplay Vulkan support

convextriangle:
Adds a Vulkan implementation of RageDisplay (with custom shader support for it).

To test, in Preferences.ini:
```
VideoRenderers=vulkan, d3d, opengl
```

Optionally, to view possible Vulkan issues (also in Preferences.ini) (note that this tanks performance):
```
VideoRendererDebugMode=1
```

Also, [Vulkan SDK](https://vulkan.lunarg.com/sdk/home) is necessary for building Vulkan stuffs (some drivers ship with older versions of Vulkan libraries, specify the SDK location in the VULKAN_SDK env var so CMake can pick it up - see [here](https://cmake.org/cmake/help/latest/module/FindVulkan.html) for info)...

Note - macOS support doesn't exist because I don't have a mac to test/code the MoltenVK support / Metal surface...

PR #1467, opened 2026-08-15
Patch up Vulkan renderer inconsistencies pt. 1

convextriangle:
- `WITH_VULKAN` CMake flag seems broken on some systems
- some things to fix (try MSAA from this): https://github.com/etternagame/etterna/pull/1418#issuecomment-5298996922
- videos get zoomed in / something something UpdateTexture broke
- some display mode info is missing (when compared w/ OpenGL & D3D9)
- swapchain creation is silly (crashes either on switching themes or on minified window)

### DIFF
--- a/.ci/generate_doc_configs.sh
+++ b/.ci/generate_doc_configs.sh
@@ -10,4 +10,4 @@ eval "$(luarocks path --bin)"
 
 # Generate Etterna CMake
 mkdir build && cd build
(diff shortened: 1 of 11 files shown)


poco0317:
vulkan renderer seems to have gained 20-60% fps with these changes maybe. not sure if its vulkan or other changes i did to network threading. im getting 4000 fps on the main menu and 2500 in gameplay

tap glow (NoGlow preference) is still missing
anti aliasing appears better

PR #1470, opened 2026-08-17
Fix pitch off downrate sync

graemephi:
Context for the fix is that the game constantly asks the RageSoundReader playing the music "where u at" to determine the current music time. This is an integer that refers back to the original music file's sample position. When you are doing unpitch rates you are mixing together multiple samples from the original stream in arbitrary ways so there is not necessarily a clean answer to this question. The way I want to think about is the stream just advances samples at a constant rate that is slower or faster than 1 sample per source stream sample. So the original code does that relative to where the source stream is parked at at the time. Why doesn't this work. I dunno

Since that doesn't work I moved it over to the next obvious thing, which is averaging all the sample positions you're using together, for every singe sample. This feels like I'm missing something, because the game doesn't ask for the position of every sample. But it works

While I was poking around in there I messed around with the algorithm for down rates a bit. Most things I tried made it sound worse. Will what's left make an audible difference?? Who can say


poco0317:
in testing this seems to make the sync problem maybe twice as worse

PR #1472, opened 2026-08-22
Fix pitch off high uprates

graemephi:
Last PR was totally busted for uprates above 2.0 👍


poco0317:
want to note that in testing i got inconsistent sync results usually tending about 20ms late but after switching between implementations several times and testing the same files over again it seems the same as before and it made no difference (not broken -> not broken, or broken -> broken). tested rates 1x-2x on some easy files including dire dire docks

PR #1482, opened 2026-09-02
Add proportional Cyrillic pages to the theme fonts

MAX-WiRED:
Cyrillic text is drawn with roughly ten pixels of dead space after every
letter, so a Russian title reads as `П л а к -П л а к` instead of
`Плак-Плак`. This adds a proportional Cyrillic page to the theme fonts so
the letters sit next to each other.

## Screenshots

Til Death, the same pack before and after. Look at `Быть дауном` and
`Плак-Плак` against the Latin and Japanese titles around them.

Before:

Снимок экрана 2026-08-24 233920

After:

Снимок экрана 2026-09-02 220523

## Why it happens

The bitmap fonts have no Cyrillic page of their own, so the letters come from
`[symbol2]`, which is row 7 of JIS X 0208. In that encoding Cyrillic is
**full-width (zenkaku)**, and the font description says so: all 66 frames are
declared 24 px wide.

In `Themes/Til Death/Fonts/bokutachinogothic2bold/doubleres/24px.ini`, the
`[symbol2]` block gives frames 96..161 - the 66 Cyrillic letters - the same
width of 24, with no exceptions.

`FontPage::SetTextureCoords` in `src/Etterna/Models/Fonts/Font.cpp` uses
that single number for two different things:

```cpp
g.m_fWidth = static_cast<float>(widths[i]);
...
g.m_iHadvance = static_cast<int>(g.m_fWidth) + iAdvanceExtraPixels;
```

The declared width is both the crop rectangle and the advance. The glyphs
themselves are 10-22 px of ink inside a 24 px full-width cell, so the cursor
moves 24 px regardless and every letter is followed by the leftover.

## Why the widths cannot simply be corrected

The crop is symmetric around the centre of the cell, while the Cyrillic glyphs
in `[symbol2]` are not centred: they sit 1 to 11.5 px left of centre, and by
different amounts per letter. Narrowing the declared widths would take the
missing pixels off both sides equally and clip the letters.

## What this changes

A separate page with centred glyphs and real advances taken from the font
metrics, added to the theme fonts that need it:

| Theme | Fonts |
|---|---|
| Til Death | `bokutachinogothic2bold` 24px/48px, `open sans semibold` 24px/48px |
| Rebirth | `_theFont` 24px/48px, `_open sans` 48px, `_open sans Bold` 48px |
| `_fallback` | `_open sans semibold` 24px, `_roboto black Bold` 24px |

The page is named `[zcyrillic]` on purpose. Font pages are loaded in sorted
filename order and the last one wins, so a name starting with `z` overrides the
full-width Cyrillic from `[symbol2]` without touching that page at all.

Coverage is U+0400-U+045F, 96 codepoints, laid out 16x6 - Russian, Ukrainian,
Belarusian, Serbian and Macedonian.

A small `[zpunct]` page comes with it for the punctuation those titles need and
the stock pages either do not map or map to a blank cell: em and en dashes,
guillemets, low quotation mark, bullet, ellipsis, curly quotes, and U+2116 (№)
where the theme font lacks it.

## Licensing

Glyphs are rendered from **Open Sans** and **Roboto**, both SIL Open Font
License 1.1. Only rendered bitmaps are added; no font file is redistributed,
and the licence line is repeated in every `.ini` section that was added.

The weight of each page was not guessed from the font name. For every target
the stem of the stock `H` was measured on that font's own `[main]` (or `[ank]`)
page, and the variable font instance whose stem matches at the same cap height
was used: SemiBold for the Til Death pages and `_theFont`, Regular and Bold for
Rebirth's menu fonts, Roboto Black for `_roboto black Bold`.

## Diff shape

- 10 `.ini` files, append-only: a commented block at the end of each, nothing
  above it is touched.
- 24 new PNG atlases, about 680 KB in total.
- No code changes.

## Testing

Applied to a 0.75.0 install and checked in the song wheel of Til Death and
Rebirth: Russian titles advance tightly and sit at the same weight as the Latin
beside them.

For the fonts that ship a `-stroke` page, the halo was matched to the stock one
by measuring how its alpha falls off with distance from the ink rather than by
eye. `_theFont` reads 184, 135, 78, 36, 9, 0 at one to six pixels out; the
generated page reads 177, 119, 70, 28, 6, 1. `_fallback` uses a hard one pixel
outline instead, and its page was built to match that.

The existing `[symbol2]` page is left in place, so Japanese titles that use
row 7 are unaffected.

## Not covered here

Some charts have the spacing baked into the file itself, with `#TITLE`
literally containing `Г О С П О Ж  А Л И С А`. No font change can help those - they read
the same way in every editor. That is a data problem in the packs, mentioned
here only so the two are not confused.

image


poco0317:
this is an AI generated pr

PR #1484, opened 2026-09-13
Fix macOS OpenGL windowing and improve Apple Silicon audio performance

Sebayop:
I ran into rendering problems when resizing or maximizing Etterna on my MacBook Pro. I found that native window changes could leave the OpenGL viewport smaller than the drawable, and the windowed VSync preference could disagree with the actual swap interval.

This patch corrects those window and context updates. It also adds missing macOS bundle metadata, enables ARM64 FFT acceleration, and works around repeated LuaJIT machine-code allocation failures in Rebirth's audio visualizer.

## Changes

- Update the macOS OpenGL context synchronously before drawing after resize, move, screen, backing-scale, and native fullscreen changes. Use the drawable dimensions and preserve the current Retina surface policy.
- Apply VSync to the main and background contexts, remove the extra frame presentation during resolution changes, and avoid resizing or recentering for VSync-only changes.
- Populate the bundle identifier (`org.etterna.Etterna`), name, and version fields through CMake.
- Enable ARM64 NEON FFT kernels through the existing sse2neon dependency, retaining scalar fallback and x86 dispatch. Add numerical regression tests and a benchmark.
- Interpret the Rebirth audio visualizer module on macOS ARM64 with `jit.off(true, true)`, while keeping other modules' existing JIT settings.
- Add a native window regression harness and instructions for running it in isolation.

## Testing

The full-game tests below were run on my local 0.75.0 build, based on `989bb7987042ffe659c7894cbd1e8628b770e2ec`. I applied the patch cleanly to current `develop` (`1abe8d8f29e7c14f514709b50671a91d61329041`) and verified that all 11 changed files match the tested versions byte for byte. I have not rerun the full-game tests on this newer base.

I tested on my MacBook Pro (`MacBookPro18,1`): Apple M1 Pro, 10-core CPU, 16-core GPU, 16 GB RAM, and macOS 26.5 (25F71). I used the built-in Liquid Retina XDR display, whose mode reported 120 Hz. The panel's native resolution is 3456 × 2234; the tested native-fullscreen OpenGL drawable was 1728 × 1084.

- I built the complete ARM64 Release configuration. All seven native window transitions passed with VSync enabled and disabled, including maximize/restore and entering/exiting Cocoa fullscreen. The original window code reproduced the viewport and swap-interval failures.
- FFT regression passed 833 numerical comparisons, including independent small DFT references and round trips. AddressSanitizer/UndefinedBehaviorSanitizer and x86_64 tests also passed. I ran x86_64 through Rosetta on this Mac. The tested 1024–8192-point audio FFTs were approximately 1.9–2.1× faster; that measures FFT execution, not game FPS.
- I checked bundle metadata with the default and repository plist templates, preserving the Retina flag.
- I tested the visualizer workaround in an isolated copy of my affected 0.74.4 installation. Repeated song changes completed without the allocation-retry stall, with global JIT still enabled. The current repository script also passed a syntax check with the game's LuaJIT.
- I measured 119.99 FPS in one autoplay chart using the complete local 0.75.0 build, native fullscreen, and a local 120 FPS cap. Over a 15-second sample, 1,782 frames had p95/p99 intervals of 8.51/8.76 ms. Screenshot readback and the following frame were excluded. This measures application presentation calls, not physical panel scanout, and covers the combined build rather than an isolated change.
- `git diff --check` passes.

## Known limitations

- I validated Cocoa native fullscreen. Legacy exclusive fullscreen using display capture remains unresolved.
- The shared `ResolutionChanged()` change also affects other OpenGL platforms. I have not tested Windows/Linux game runs, multi-monitor transitions, or other Mac configurations.
- Bundle metadata does not grant Input Monitoring access. Changing an ad hoc signed executable can require permission to be granted again.
- The visualizer change is a scoped LuaJIT workaround. My 120 FPS preferences and installation recovery are local settings, outside this patch.


poco0317:
this is a suspicious pr because your overview seems to be put through AI to translate and clean it up.
you are modifying the external dependencies which we prefer not to do because we dont own that code and cant maintain the changes when we update the libraries.
you are modifying the windows/linux opengl flows without testing it for the sake of an os barely anyone uses. it needs testing and effort. 
the lua change should not be necessary on the version you tested due to the luajit updates, and i have no way to prove one way or the other.

usually adding testing for no clear reason is a strong sign of ai generated code, but one thing that saves you here is that it looks like someone who codes shit to get things done did it.

if you have used to ai to make this code change, it is required that you tell us so that i can just close and forget about the pr

Sebayop:
Yes, I did use codex 

I originally made this patch for my own local instance just to get things working so I could play normally. I only opened the PR to share the solution I'm using locally, in case there was anything useful in it that the maintainers could take from it.

What happens with the PR is entirely up to you. If you prefer to close it because of that, I completely understand and respect the decision.

